PR #258, opened 2014-07-02
netbase: Swapping hostname and ip in the /etc/hosts file.

sharidas:
Swapping the hostname and ip in the /etc/hosts file, appended
in this bbappend file.

JIRA: SB-3004

Signed-off-by: Sujith H Sujith_Haridasan@mentor.com


kergoth:
Hmm, I was looking at submitting this upstream, but this shouldn't be needed, as far as I can tell. systemd provides its own version of nss-myhostname, so our host name should already resolve via that, unless our nsswitch.conf isn't configured to use it. I'll be investigating that approach rather than explicitly adding a hosts entry for upstream.

PR #269, opened 2014-07-22
alsa-lib: Properly handle slave xrun conditions for DSHARE/DSNOOP/DMIX

mdurnev:
If using very short periods, DSHARE/DSNOOP/DMIX may report underruns while in
status 'prepared'. This prohibits correct recovery.

Update alsa-lib to properly handle slave xrun conditions for DSHARE/DSNOOP/DMIX

JIRA: ADK-3329

Signed-off-by: Mikhail Durnev <mikhail_durnev@mentor.com>


kergoth:
This was against an old release. Presumably it belongs in a hotfix for folks running it, otherwise it can be applied to newer releases if still needed. @mdurnev, what should be done with this? Does it still need merging to the old release branch?

PR #408, opened 2015-01-14
libunwind: fix linker error by linking against libunwind library

fahadarslan:
From 4c62c4a9556f92495d93eddc7641497e1ce1e35c Mon Sep 17 00:00:00 2001
From: Ulrich Weigand <uweigand@de.ibm.com>
Date: Tue, 17 Dec 2013 15:00:54 +0100
Subject: [PATCH] Fix test case link failure on PowerPC systems with Altivec

Upstream-Status:backport

On systems where the system compiler supports Altivec by default,
the libunwind Makefile will attempt to build an extra test case
ppc64-test-altivec.  Unfortunately, the link step will fail since
the Makefile does not actually link against the libunwind library.

Fixed by adding the appropriate LDADD macro.

Signed-off-by: Ulrich Weigand <uweigand@de.ibm.com>

### DIFF
--- a/meta-mel/classes/qoriq_build_64bit_kernel.bbclass
+++ b/meta-mel/classes/qoriq_build_64bit_kernel.bbclass
@@ -1,0 +1,18 @@
+inherit distro_features_check
+REQUIRED_DISTRO_FEATURES_e6500 += "multiarch"
+
+python () {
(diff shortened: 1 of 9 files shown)


Noor-Ahsan:
Looks good to me.

PR #412, opened 2015-01-23
Sb 4374 layerscape bsp

arun-khandavalli:


arun-khandavalli:
@kergoth : Can you please help in reviewing the PR

meta-mel/fsl-arm/recipes-bsp/u-boot/u-boot-ls1_%.bbappend
```diff
@@ -0,0 +1,2 @@
+UBOOT_CONFIG = "sdcard"
```
kergoth: I'm fairly sure doing this in the recipe is wrong. This is a configuration variable which is generally set in the machine, not the recipe. I think we should use a local.conf fragment to add `UBOOT_CONFIG_ls1021atwr = "sdcard"`, the same way we do for other similar configuration changes for upstream machines like minnow (see [meta-mel/conf/local.conf.append.minnow]()).

meta-mel/fsl-arm/recipes-core/images/console-image.bbappend
```diff
@@ -0,0 +1 @@
+IMAGE_INSTALL += " u-boot"
```
kergoth: Why would we want u-boot installed in the rootfs? Does the early stage bootloader know how to load it from there? This isn't generally how things are done, in my experience.

skrishnakar:
@arun-khandavalli Please move UBOOT_CONFIG to local.conf.append.ls1021atwr as suggested by Chris. 

Also add this change into same machine append file, for getting u-boot built by default:

EXTRA_IMAGEDEPENDS += "virtual/bootloader"

arun-khandavalli:
@skrishnakar  @kergoth : As suggested I have made the changes as suggested. Please let me know for any further changes.

@kergoth: Actually, It helps in having u-boot built, Board on power-up internal bootloader will load 1st and it will read the BOOT-MODES which are hardware registers,  so depending on boot-mode the i-boot will load the secondary boot-loader u-boot ( universal boot-loader).

SW2.1: OFF, SW2.3: ON - SD-Boot,  u-boot shall be loaded from SDCARD.
SW2.1: ON , SW@2.3 OFF- NOR-Boot. u-boot shall be loaded from NOR.

kergoth:
IMAGE_INSTALL is used to install packages into the image. As @skrishnakar points out, EXTRA_IMAGEDEPENDS Is how you add build time dependencies to images, rather than runtime.

meta-mel/conf/local.conf.append.ls1021atwr
```diff
@@ -0,0 +1,2 @@
+UBOOT_CONFIG_ls1021atwr = "sdcard"
+EXTRA_IMAGEDEPENDS += "virtual/bootloader"
```
kergoth: This should be `EXTRA_IMAGEDEPENDS_append_ls1021atwr = " virtual/bootloader"`, and we should likely add comments above these lines indicating their purpose to clarify for the user, since this will end up visible to them in their local.conf. Again see the minnow one for an example.

arun-khandavalli:
The changes of this PR has been merged through different PR. Hence closing this request

PR #414, opened 2015-01-26
mel: remove toolchain from Yocto SDK for external

kergoth:
As we're using an external toolchain, we don't want to ship a newly built
toolchain inside a Yocto SDK.

This is an _remove, becuase I didn't want to hardcode the full package list
for TOOLCHAIN_HOST_TASK. The problem with this is, once an _remove is defined,
there's no way to undo it, so a user wouldn't be able to undo this. It's
placed in a local.conf.append to ensure it's user visible, letting them
comment it out or delete the line to undo it if they choose to.

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -121,6 +121,10 @@ SSTATE_MIRROR_SITES += "file://${MELDIR}/cached-binaries"
 # The default value follows:
 #TMPDIR = "${TOPDIR}/tmp"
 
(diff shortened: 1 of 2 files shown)


kergoth:
@Noor-Ahsan, @challinan

kergoth:
@Noor-Ahsan, fixed. I just noticed that I already had the cross canadian packagegroup removed in meta-sourcery, though, so will submit a pull req there to drop the insufficient change from that layer in favor of this.

PR #417, opened 2015-01-30
prepare-boot-media.sh: adapt the script for p1020/p2020rdb

fahadusman:
Made the script independent of machine name, now user would have to specify
the complete image name with path while invoking the script.


Noor-Ahsan:
does it work on prebuilt binaries as well?

fahadusman:
yes, but the user would have to invoke it differently, specifying the file-name of image file present in pre-built images.

PR #421, opened 2015-02-10
ls1021atwr: Enable unionfs support for the LS1

arun-khandavalli:
Union-Fs support for the LS1 has been enabled and verified on
the target.

Also corrected right way of appending the kernel config fragments,
else would result in compilation errors.

JIRA:SB-4334.


arun-khandavalli:
@skrishnakar : Can you please help in reviewing the patchset, added the unionfs feature and corrected the right way of appending the kernel config fragments.

skrishnakar:
Looks reasonable !

PR #431, opened 2015-03-02
LS1021A: Install Alsa package through Machine features

arun-khandavalli:
Alsa/Audio packages should not be installed directly, Corrected the
installation of Alsa into image using the machine features.

JIRA:SB-4642

### DIFF
--- a/meta-mel/conf/local.conf.append.ls1021atwr
+++ b/meta-mel/conf/local.conf.append.ls1021atwr
@@ -1,0 +1,2 @@
+# LS1021atwr supports usb-audio & simple soundcard, alsa utils are required for playback/record operations
+MACHINE_FEATURES_append_ls1021atwr = " alsa"
--- a/meta-mel/fsl-arm/recipes-core/images/console-image.bbappend
+++ b/meta-mel/fsl-arm/recipes-core/images/console-image.bbappend
@@ -1,1 +1,0 @@
-IMAGE_INSTALL += "packagegroup-tools-audio"


arun-khandavalli:
@kergoth @skrishnakar @Noor-Ahsan : Can you help in reviewing the new patch-set, As per @kergoth  comments corrected the installation of Alsa package through machine feature.

meta-mel/conf/local.conf.append.ls1021atwr
```diff
@@ -0,0 +1 @@
+MACHINE_FEATURES_append_ls1021atwr += "alsa"
```
kergoth: Please don't combine _append and +=, they're both append operations. Instead, make the separator explicit:
```
MACHINE_FEATURES_append_ls1021atwr = " alsa"
```
Also, we should add a comment above this. It will show up in the user's local.conf, so we should at least make a best effort to explain why we're overriding/altering things the machine normally takes care of at that level.

Other than that, this looks great, thanks.

arun-khandavalli:
@kergoth : As suggested, made the changes accordingly. Please help in reviewing the patchset2.

meta-mel/conf/local.conf.append.ls1021atwr
```diff
@@ -0,0 +1,2 @@
+#LS1021atwr supports usb-audio & simple soundcard, alsa utils are required for playback/record operations
```
kergoth: I hate to nitpick here, but this doesn't line up with all the other comments in the file. We wrap long lines, and use a space after # :)

arun-khandavalli:
Thanks @kergoth , updated the new patch-set with latest review.

PR #436, opened 2015-03-06
kernel-host-dev: pass required flags to build for t4240rdb-64b

fahadarslan:


kergoth:
What's the impact of this on other machines? Is that bbclass always available?

fahadarslan:
Originally this class is being provided by meta-fsl-ppc. I've placed slightly modified version of this at ~/meta-mentor/meta-mel/classes/qoriq_build_64bit_kernel.bbclass. So my understanding is it should be always available and shouldn't impact other machines since it returns if ppc64e6500 is not found in TUNE_PKGARCH.

PR #437, opened 2015-03-10
mel-lite.conf: fix validated host distro

abelal:
Validated host for mel-lite 2014.12 is only Ubuntu 14.04
so we make the adjustment accordingly.

### DIFF
--- a/meta-mel/conf/distro/mel-lite.conf
+++ b/meta-mel/conf/distro/mel-lite.conf
@@ -7,5 +7,5 @@ DISTROOVERRIDES = "${DISTRO}:mel"
 require conf/distro/include/mel.conf
 
 SANITY_TESTED_DISTROS = "\
-    Ubuntu-12.04 \n\
+    Ubuntu-14.04 \n\
 "


abelal:
@drewmoseley, please review.

drewmoseley:
Acked-by: Drew Moseley \<drew_moseley@mentor.com\>

@Noor-Ahsan is this branch open for checkin?

abelal:
@Noor-Ahsan, ping!

Noor-Ahsan:
@drewmoseley , @abelal  yes this branch is open for bug fixes. I have one question here. Why you are not using MEL SANITY DISTROes and using your own. We define this variable in mel.conf and I see you are including that file as well.

abelal:
@Noor-Ahsan, the answer is probably that we limit the validation costs on mel-lite for AMD but let's hear it from @drewmoseley.

drewmoseley:
That's correct.  For MEL Lite we opted to only test specifically on one host since it's a free download.  We are only testing on Ubuntu 14 for MEL Lite although we fully expect other hosts to work.

@Noor-Ahsan ok to merge?

Noor-Ahsan:
I am fine with this change But how you will do it in future in case MEL repo is not taking anymore changes?

drewmoseley:
We can override it in meta-amd if we need to.  In fact we did that previously and this is just the cleanup to move it here.

PR #438, opened 2015-03-12
mel.conf: Enable vfat support for ARM targets

arun-khandavalli:
Add 'vfat' to MACHINE_FEATURES which in turn includes
dosfstools via packagegroup-base

JIRA: SB-4678


arun-khandavalli:
@Noor-Ahsan @skrishnakar : Can you please help in revewing the patch

Noor-Ahsan:
@arun-khandavalli  this should be in the machine file instead of mel.conf. We did it for power because we don't own ppc machine files.

PR #439, opened 2015-03-12
udev-extraconf: Adding modified version of mount.sh

sharidas:
Adding modified version of mount.sh to handle devices
without label. Its taken care with the help of UUID,
using blkid command.

JIRA: SB-4637

Signed-off-by: Sujith H <Sujith_Haridasan@mentor.com>


sharidas:
@kergoth I had modified the mount.sh from udev-extraconf, to get device mounted with label id's.

PR #443, opened 2015-03-19
connman 1.25 -> connman 1.28

fahadusman:
Move to connman 1.28 plus a couple of extra patches to address the network issues reported in SB-4740.


kergoth:
New recipes are not appropriate for a distro layer like meta-mel, they belong in meta-mel-support.

PR #444, opened 2015-03-20
connman 1.25 -> connman 1.28 (v2)

fahadusman:
Move to connman 1.28 plus a couple of extra patches to address the network issues reported in SB-4740.


meta-mel-support/recipes-connectivity/connman/connman_1.28.bb
```diff
@@ -0,0 +1,13 @@
+require recipes-connectivity/connman/connman.inc
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
```
kergoth: This is incorrect. FILESEXTRAPATHS is a mechanism provided to bbappends. The recipe already knows to look at the files dir next to the recipe. Just drop this line.

PR #448, opened 2015-03-30
os-release: add to SIGGEN_EXCLUDERECIPES_ABISAFE

kergoth:
We don't want systemd and everything that depends on it rebuilding whenever
DISTRO_VERSION changes.


kergoth:
This is a must for sstate reuse.

PR #450, opened 2015-03-31
watchdog: kill update-rc.d when not using sysvinit

kergoth:
Fixes a dependency on a nonexistent initscripts-functions package.


kergoth:
This was breaking the p1010rdb build for me with updated layers.

PR #456, opened 2015-04-21
bb-determine-layers: show paths for duplicated layers

kergoth:
JIRA: SB-3329

### DIFF
--- a/scripts/bb-determine-layers
+++ b/scripts/bb-determine-layers
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 
-import contextlib
 import glob
 import optparse
 import os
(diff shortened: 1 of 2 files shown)


kergoth:
Example output after this is applied:
```
› . ./meta-mentor/setup-environment -f qemux86
You had no local.conf file. This configuration file has therefore been
created for you with some default values.
You had no bblayers.conf file. This configuration file has therefore been
created for you with some default values.
Locating layers and parsing layer.conf files...done
Resolving layer dependencies...done
Determining layers to include for MACHINE 'qemux86'...failed (duplicate layers)
Multiple layer paths found for layer name 'core':
  /scratch/test/openembedded-core/meta
  /scratch/test/poky/meta
Error in execution of bb-determine-layers, aborting
```

kergoth:
An additional improvement I'd like to see done after this would be to let the user specify the layer filesystem path via -l to select which of the duplicates they prefer, rather than having to remove one from disk. E.g. in this case: `. ./meta-mentor/setup-environment -f -l poky/meta -f qemux86`

PR #457, opened 2015-04-22
Adjust for layer updates

kergoth:


kergoth:
Depends on MentorEmbedded/mel-manifests#153

PR #465, opened 2015-05-15
Include fsl-networking by default for ls1021atwr

kergoth:
JIRA: SB-5077


kergoth:
I'm not 100% happy with this, but at least it avoids hardcoding the conditionals in the setup scripts themselves.

PR #470, opened 2015-05-22
docbook-utils-native: build fix by static linking openjade

sbobade:
Openjade has an issue being built on x86_64 host
due to a shared lib linking issue (libogrove.so),
which fixed by using static linking.

reference:
https://bugzilla.yoctoproject.org/show_bug.cgi?id=2972#c0
http://pastebin.com/cUb1x2Qj

JIRA: SB-5027

Signed-off-by: Neeraj Sharma <neeraj.sharma@alumni.iitg.ernet.in>
Signed-off-by: Shrikant Bobade <Shrikant_Bobade@mentor.com>

### DIFF
--- a/meta-mentor-staging/recipes-devtools/openjade/openjade-native_%.bbappend
+++ b/meta-mentor-staging/recipes-devtools/openjade/openjade-native_%.bbappend
@@ -1,0 +1,23 @@
+EXTRA_OECONF += "--enable-static --disable-shared"
+
+# override do_install
+do_install() {
+        # Refer to http://www.linuxfromscratch.org/blfs/view/stable/pst/openjade.html
+        # for details.
+        install -d ${D}${bindir}
+        install -m 0755 ${S}/jade/openjade ${D}${bindir}/openjade
+        ln -sf openjade ${D}${bindir}/jade
+
+        install -d ${D}${datadir}/sgml/openjade-${PV}
+        install -m 644 dsssl/catalog ${D}${datadir}/sgml/openjade-${PV}
+        install -m 644 dsssl/*.dtd ${D}${datadir}/sgml/openjade-${PV}
+        install -m 644 dsssl/*.dsl ${D}${datadir}/sgml/openjade-${PV}
+        install -m 644 dsssl/*.sgm ${D}${datadir}/sgml/openjade-${PV}
+
+        install -d ${datadir}/sgml/openjade-${PV}
+        install -m 644 dsssl/catalog ${datadir}/sgml/openjade-${PV}/catalog
+
+        install -d ${D}${sysconfdir}/sgml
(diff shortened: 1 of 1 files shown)


sbobade:
@Noor-Ahsan @kergoth @skrishnakar 

Please review and merge. 

Thanks
Shrikant

kergoth:
This commit message is just not right. It implies docbook-utils-native is being modified, such is not the case. Also, this commit message implies that only docbok-utils-native is affected, but this is wrong, anything linking against openjade-native will link against it statically. Further, the do_install override seems wrong. Why are we installing everything manually rather than letting autotools do it? If the upstream make install target can't handle --disable-shared, then it should be fixed to do so.

Do we know if everything that links against openjade-native is affected by this, or just dobook-utils-native? If the latter, I'd recommend just altering the docbook-utils-native build to link openjade statically but everything else shared, e.g. -Wl,-Bstatic -lopenjade -Wl,-Bdynamic, if at all possible.

Also, you don't mention gold at all in the commit message, but afaict this is specific to linking with gold on the host.

sbobade:
@kergoth 

yes, I will update the commit message..as per comments and this issue is specific to gold linker so will include gold details in commit message.
Please have a look on below investigation points  as per review comments.. detail logs ref. updated on tracking jira

i. currently w.r.to core-image-sato docbook-utils-native is only consumer for openjade-native
so we observed the build break w.r.ro docbook-utils-native only. 
ii. we have analysed w.r.to gold linker and bfd linker, 
observed  libogrove.so not got linked to openjade properly in case of gold linker, 
while in case of bfd linker it got linked successfully and build got successful.

As per recommendation tried linking only lib openjade statically with docbook-utils-native but actually openjade-native not got linked to local libs properly results in build break.

working on these two ways to prepare the fix.. 
1. Openjade: link local libs statically, so in presence of gold linker there will be no build break of docbook-utils-native [ checking with cleaner approach].
changes :  we need to modify the upstream make install (do_install) to remove the code related to shared libs in order compile openjade-native with only static local libs successfully : --enable-static --disabled-shared, In this case only openjade-native will be updated and not docbook-utils-native [changes mostly similar to the above patch of this PR]
2. Explicitly trying to enforce use of bfd linker , similar to LD = "${HOST_PREFIX}ld.bfd "
But currently it is in progress I am getting build breaks..looking into it.

Please advice/suggest changes or modifications if any.  

Thanks
Shrikant

sbobade:
@Noor-Ahsan @kergoth @skrishnakar 

Please review and merge updated PR.
openjade-native: statically linking its local libs to fix the dynamic linking issue specific with gold linker (ld.gold). 
Please suggest changes or modifications if any, 

Thanks
Shrikant

kergoth:
Looks okay for now. ld.bfd use might be cleaner in the long term, and we'll want to push this upstream, but this will get the job done in the meantime.

PR #472, opened 2015-05-28
openjade-native: fix link issue with ld.gold by static link

sbobade:
This patch provide static link of openjade with its local libs,
it is required to fix the link issue specific with gold linker.

In case of gold linker we unable to get dynamic link with local lib:
libogrove.so, which results in build break of its consumers.

Overriding do_install to handle the disable shared libs.

reference:
https://bugzilla.yoctoproject.org/show_bug.cgi?id=2972#c0
http://pastebin.com/cUb1x2Qj

JIRA: SB-5027

Signed-off-by: Neeraj Sharma <neeraj.sharma@alumni.iitg.ernet.in>
Signed-off-by: Shrikant Bobade <Shrikant_Bobade@mentor.com>

### DIFF
--- a/meta-mentor-staging/recipes-devtools/openjade/openjade-native_%.bbappend
+++ b/meta-mentor-staging/recipes-devtools/openjade/openjade-native_%.bbappend
@@ -1,0 +1,23 @@
+EXTRA_OECONF += "--enable-static --disable-shared"
+
+# override do_install
+do_install() {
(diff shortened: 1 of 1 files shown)


sbobade:
@Noor-Ahsan  @kergoth @skrishnakar 
Please review and merge,

Thanks
Shrikant

PR #474, opened 2015-06-04
pulseaudio: make starting up PA deterministic

abelal:
The autostart behavior for PA is fixed with the help of some patches from upstream. One additional patch that strictly targets MEL is to disable the auto-spawning feature of PA so it is completely controllable through systemd.


abelal:
@Noor-Ahsan, @drewmoseley please review.

kergoth:
Making it behave well with systems is clearly of value in general, rather than being mel specific. Please move to meta-mentor-staging.

kergoth:
Systemd, that is.

abelal:
@kergoth, I have reworked the PR as suggested but I have not moved one change (disabling autospawn by default) to staging because that is MEL specific in my opinion.

abelal:
@Noor-Ahsan, @kergoth, ping!

PR #479, opened 2015-06-24
mel-versions.conf: use connman 1.28 for AMD platforms

abelal:
The AMD platforms utilize connman 1.28 when built on
mel or mel-lite so we set the version generically rather
than from each BSP layer.


abelal:
@drewmoseley, @Noor-Ahsan, please review.

PR #482, opened 2015-06-29
Add eventmask to some event handlers

kergoth:
This fixed builds with current layers, though upstream has since fixed the
code to be more tolerant. Even so, event handlers should all be using
eventmask anyway to avoid their running for events they don't care about.


Noor-Ahsan:
@kergoth  your signeoff need an update :).

PR #484, opened 2015-07-01
local.conf.sample: disable bluetooth support for qoriq-ppc

arehman199:
JIRA: SB-5157

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -81,6 +81,9 @@ USER_FEATURES += "~3g"
 # Disable near field communication support by default
 USER_FEATURES += "~nfc"
 
+# Disable bluetooth support for qoriq-ppc
+USER_FEATURES_qoriq-ppc += "~bluetooth"
+
 # How many tasks bitbake should run in parallel
 #BB_NUMBER_THREADS = "4"
 


kergoth:
First of all, you're disabling it for qoriq, not all of the powerpc architecture, so the commit message subject is not accurate. Second, your commit message doesn't indicate *why* this is necessary. Commit messages should always include not just what is being done, but why it's being done, unless it's self evident, which it's not here.

arehman199:
Commit updated.

kergoth:
I'd like a bit of clarification here. Why is the bluetooth service failing on this platform? If the hardware doesn't support bluetooth, then why does the MACHINE_FEATURES for those machines include it? Only if both MACHINE and DISTRO features include bluetooth will it be included in the images by default.

arehman199:
bluetooth is being included in DISTRO_FEATURES in poky/meta/recipes-core/packagegroups/packagegroup-base.bb:

> RDEPENDS_packagegroup-base-extended = "\
>     packagegroup-base \
> ...
>     ${ADD_BT} \
> ...
>     "
> 
> ADD_WIFI = ""
> ADD_BT = ""
> ...
> 
> python __anonymous () {
> 
> ...
> 
>     if "bluetooth" in distro_features and not "bluetooth" in machine_features and ("pcmcia" in machine_features or "pci" in machine_features or "usbhost" in machine_features):
>         d.setVar("ADD_BT", "packagegroup-base-bluetooth")
> 
> ...
> 
> }

We are reworking the solution.

kergoth:
[**@arehman199**](https://github.com/arehman199) commented on [Jul 3, 2015, 6:21 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/484#issuecomment-118348640 "2015-07-03T13:21:05Z - Replied by Github Comment Enhancer"):
> bluetooth is being included in DISTRO_FEATURES in poky/meta/recipes-core/packagegroups/packagegroup-base.bb:
> 
> > RDEPENDS_packagegroup-base-extended = "\
> >     packagegroup-base \
> > ...
> >     ${ADD_BT} \
> > ...
> >     "
> > 
> > ADD_WIFI = ""
> > ADD_BT = ""
> > ...
> > 
> > python __anonymous () {
> > 
> > ...
> > 
> >     if "bluetooth" in distro_features and not "bluetooth" in machine_features and ("pcmcia" in machine_features or "pci" in machine_features or "usbhost" in machine_features):
> >         d.setVar("ADD_BT", "packagegroup-base-bluetooth")
> > 
> > ...
> > 
> > }
> 
> We are reworking the solution.

No, bluetooth is added to distro_features by the distro. packagegroup-base is what includes the bluetooth packages when it's in both machine and distro features, which I already explained. You haven't answered my original question: why is bluetooth in the machine features?

arehman199:
Actually bluetooth is not in the MACHINE_FEATURES, it is only in the DISTRO_FEATURES. The reason it is getting part of the image is due to the check in anonymous python function. This function is what is responsible for populating the ADD_BT variable with bluetooth packagegroup. 
As per my understanding, this function adds bluetooth package even if it is not in the machine_features but the interfaces (pci, pcmcia, usbhost) that support external bluetooth devices are available. So the reason the service is getting failed on target is due to the absence of external bluetooth device on such interfaces. After discussing it with @Noor-Ahsan , we've decided to remove the ADD_BT variable from packagegroup-base-extended altogether for ppc targets.

arehman199:
@Noor-Ahsan @kergoth Kindly review: https://github.com/MentorEmbedded/meta-mentor/pull/489

PR #491, opened 2015-07-13
sstate-mirror-sites: fix lack of needed expansions

kergoth:
If SSTATE_MIRROR_DISTROS or SSTATE_MIRROR_SITES used inline python, the
generated SSTATE_MIRRORS could have invalid syntax. Fix by using expanded
versions of those variables. Also add a check to skip empty sites.


kergoth:
This is needed to fix mel build failures with current metadata, so merging immediately.

PR #492, opened 2015-07-13
imx-gpu-viv: correct file ownership

kergoth:
This recipe copies binary files with `cp`, including `cp -a`, so the file
ownership can end up being the same as the build user. Work around this by
chown'ing everything to root:root.

This avoids a package_qa failure.

JIRA: SB-5184


kergoth:
Merging now, as it resolves an mx6q build failure and is clearly trivial.

PR #494, opened 2015-07-24
Fix various problems with deploy-license-manifest

kergoth:


otavio:
The fixes are nice but it will run even without license class being in use. I think it needs to be appended to the original task so it is run only if license handling is in use.

kergoth:
That's a good point, though it's unconditionally prepended to ROOTFS_POSTPROCESS_COMMAND in license.bbclass. Also if BUILD_IMAGES_FROM_FEEDS is set, the license function will run, but no manifest will be emitted. I think I'll alter the functions to only run if the license manifest file exists.

PR #496, opened 2015-07-27
setup-environment: process updates from post scripts

joeythesaint:
Each layer may contain a post-setup-environment script that can modify,
among other things, the build directory's setup-environment script.  The
script is sourced after being generated, but isn't sourced again following
the processing of post-setup-environment.  Adding in a second source (and
dumping the initial one to /dev/null so the same message doesn't appear
twice, adds minimal overhead and allows post-setup-environment scripts an
opportunity to make changes to setup-environment (eg. to append to
environment variables such as MANPATH).

Signed-off-by: Joe MacDonald <joe_macdonald@mentor.com>


darknighte:
I will defer to @kergoth, but this looks fine to me.

Acked-by: Sean Hudson

Noor-Ahsan:
@joeythesaint  you want this to be merged for Birch release?

joeythesaint:
@Noor-Ahsan Sure, if the ship hasn't sailed on that yet.

PR #499, opened 2015-08-05
u-boot-ls1: update the licence and md5 info for u-boot ls1

arun-khandavalli:
For cedar release u-boot 2014.01 version shall be used because
with the latest release u-boot and gcc-4.9 rcu stall crash is still
observed, also updated the license infomation with the latest
sdk-v1.8

Jira: SB-5436


arun-khandavalli:
@skrishnakar  @Noor-Ahsan : can you please help in reviewing the commit

PR #501, opened 2015-08-07
u-boot-ls1: Add support for lpuart u-boot

arun-khandavalli:
lpuart u-boot is required for the the DCU/HDMI  support in the
layerscape scenario, if we want to use hdmi we need to boot
from LP uart rather than normal uart then we can have console-images +hdmi.

We need to pass console as console=ttyLP0,115200 hdmi as parameter for hdmi
support

Jira:SB-5450


arun-khandavalli:
@skrishnakar @Noor-Ahsan Can you please help in reviewing the commit.

PR #506, opened 2015-08-13
Improve our pulseaudio bits

kergoth:


kergoth:
Updated to fix the accidental removal of the other pulseaudio module (we may want to revisit whether any of those are upstream-appropriate in the future, though.. while esound makes no sense being default, i could see an argument for the port autoswitch) and also switched to using packageconfig for root autospawn control, as I think that should go upstream in some form, and it means that both of our preferences for pulseaudio behavior can be controlled in the same place in mel.conf, rather than partly in the append and partly in the distro.

PR #508, opened 2015-08-21
ls1021atwr: Add IMAGE_BOOT_FILES which will be used by wic tool

arun-khandavalli:
Wic tool reads IMAGE_BOOT_FILES which wll make up the boot-partition
of the sd-card. In the layerscape case it contains kernel image
dtb, uboot & rcw for both lpuart and general configuration.

Jira: SB-5407

### DIFF
--- a/meta-mel/conf/local.conf.append.ls1021atwr
+++ b/meta-mel/conf/local.conf.append.ls1021atwr
@@ -7,3 +7,13 @@ EXTRA_IMAGES_ARCHIVE_RELEASE = "rcw boot"
 # Do not build/deploy pulseaudio for networking board TWR-LS1021A
 DISTRO_FEATURES_remove = "pulseaudio"
 RDEPENDS_packagegroup-base-bluetooth_remove = "pulseaudio pulseaudio-server"
+
+# Add file names to go into Boot directory of Sd-card used by wic
+IMAGE_BOOT_FILES =  "uImage\
+                     uImage-ls1021a-twr.dtb\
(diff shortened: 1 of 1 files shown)


arun-khandavalli:
@Noor-Ahsan @skrishnakar : Can u please help in reviewing the commit, the kickstart file used for the layerscape will be from poky "sdimage-bootpart". Boot will be of vfat type and rootfs ext4 type.

kergoth:
Hardcoding the versions of the u-boot files seems potentially problematic, what happens if that version changes? Can we use a wildcard for that, or will we have to keep that up to date?

Other than that, this seems reasonable to me.

fahadusman:
I am not sure about Layerscape, but for most of the other targets we have symlinks in the deploy directory that point to the u-boot binary and does not contain the version string. We should be able to use that.

arun-khandavalli:
@Noor-Ahsan @skrishnakar : taken care of review comments, can u please help in review and merge.

skrishnakar:
@arun-khandavalli Can't we use "install -m 0755" instead of "cp" for these binaries ?

arun-khandavalli:
@skrishnakar thanks for the review updated the changes/ Please help in review and merge.

fahadusman:
Hi Arun, I think it would be better to split the chances into two separate commits, other than that it looks good to me.

Thanks.

arun-khandavalli:
Hi Fahad, I have made the changes, can you please review and merge.

PR #510, opened 2015-08-21
archive-release: fix few bashism issues

abelal:
The dash shell does not support usage of '==' for
string comparisons, doing so results in erronous
outcomes.
We now fix this by using a single '=' which is
supported across different shells.


abelal:
@Noor-Ahsan, please review.

PR #512, opened 2015-08-24
Pulse audio changes

arun-khandavalli:
Pulse audio changes for LayerScape BSP because of which we are seeing a hang in the core-image-sato boot of LS1.


arun-khandavalli:
@Noor-Ahsan @skrishnakar : Can u please help in reviewing of the patchset.

PR #517, opened 2015-08-25
u-boot-ls1: Deploy u-boot for lpuart and nor images

arun-khandavalli:
Install u-boot images for lpuart and nor images which shall
and also remove support for sd-card, since in latest u-boot
we can write directly to Nor from nor u-boot

Jira:SB-5407


arun-khandavalli:
@Noor-Ahsan @skrishnakar : Can u please help in review and merge

PR #522, opened 2015-08-27
tcf-agent: Remove the patch which is not applicable on latest source

arun-khandavalli:
tcf client patch was added in previous release
to resolve the "LDFLAGS warning during QA_PACKAGE stage" but it
is not now applicable on the latest source hence removing the
patch and the bbappend.


arun-khandavalli:
@Noor-Ahsan @skrishnakar : kindly review and merge.

skrishnakar:
Looks reasonable.

Noor-Ahsan:
do you get QA warning when this patch is removed or it has been fixed in newer version?

arun-khandavalli:
@Noor-Ahsan : This patch is not applicable on the latest source, it dosent apply cleanly and it breaks the build.

Noor-Ahsan:
@arun-khandavalli that I understood. I was asking do we need similar kind of patch to fix QA warning or the new version is not showing the warning.

arun-khandavalli:
@Noor-Ahsan : Yes we need a new patch to fix the QA-warning in the latest source.

PR #537, opened 2015-09-11
connman: use PACKAGECONFIG to control systemd option

arehman199:
Conditionally add systemd to PACKAGECONFIG as in the original recipe


arehman199:
props to @Yasir-Khan for providing the fix

PR #539, opened 2015-09-12
weston-image: introduce new graphics image recipe

fahadarslan:


kergoth:
Hmm, I wonder if we should actually include/require core-image-weston, to make it clear where we differentiate from it, or if copying it is better.

fahadarslan:
@kergoth: we were also thinking which way to go. Since in case of console-image copying is being done, we took same approach.

PR #540, opened 2015-09-14
firmware-wireless: fix firmware installation path

abelal:
Firmware binaries are not architecture specific and
we inherit allarch for the same reason. In this case
we cannot use architecture specific paths such as
base_libdir which makes the firmware search fail
when it is required. To be clear the current approach
would be incorrect on 64bit systems.
We now use a non architecture specific path to fix
this issue.


abelal:

PR #541, opened 2015-09-14
Fix do_configure of libtdb and talloc

arehman199:
* Fix following error seen in configuration of libtdb and talloc
  for e500v2 targets:
  wafsamba/wscript:296: error: the configuration failed

* JIRA: SB-5793


kergoth:
Looks good to me.

PR #546, opened 2015-09-17
Archive sub-layers of mx6 separately, but keep the structure

kergoth:


kergoth:
This is fixed now, and should be good to go.

PR #548, opened 2015-09-18
mel: fix shadow build failure in non-GPLv3 builds

kergoth:
When building with GPLv3 as INCOMPATIBLE_LICENSE, the target gettext headers
will not be available, only native, yet only gettext-minimal-native provides
the native headers, which is normally pulled in by gettext-native, and
aclocal_copy_aclocals halts target->native deps to only include direct
dependencies, so we need gettext-minimal-native explicitly included as well.


kergoth:
I'm currently investigating to ensure my analysis was correct here, and to determine why the Yocto autobuilder isn't experiencing this, so we can get whatever fix is needed upstream.

kergoth:
I did manage to reproduce this, though intermittently, with just poky + meta-sourcery, in an attempt to eliminate mel bits from consideration. Continuing investigation.

kergoth:
Having re-read the code repeatedly and rewritten part of it, I can say that my initial analysis of the problem was incorrect, so I'm working to correct my mental model now.

PR #549, opened 2015-09-18
systemd: arm: disable -O2 optimization when built with GCC-5.x

skrishnakar:
There are systemd unit failures observed on multiple platforms
when built with GCC-5.x toolchain, bug reported in yocto:

https://bugzilla.yoctoproject.org/show_bug.cgi?id=8291

Workaround Will remain here until we get fix for #8291 bug.


skrishnakar:

Thanks!

PR #553, opened 2015-09-21
kmod: fix return code

arehman199:
Fix the following error seen during boot:
"Failed to insert module 'kdbus': Function not implemented"

See https://github.com/systemd/systemd/issues/203 for details

JIRA: SB-5927


kergoth:
It makes absolutely zero sense that this would go in meta-mel. How exactly is this specific to the mel distro? Why would this not go upstream to oe-core, when it's already upstream in kmod/systemd?

Noor-Ahsan:
@kergoth this will be fixed in next version. I thought when we start working on new version it will be fixed and community does not need our fix at that time. That is why I asked Abdur Rehman to send PR against meta-mel. If you disagree then let us know. We will move it to meta-mentor-stagging.

kergoth:
meta-mentor-staging is for anything which is bound for upstream, whether we're the ones sending it there or not, whether it'll be merged in the current state or not. meta-mel is generally for changes which impose distro policy or otherwise are bound specifically to how we do things and are not appropriate for upstream for some specific reason.

PR #558, opened 2015-09-25
Adjust for layer updates

kergoth:


kergoth:
This depends on https://github.com/MentorEmbedded/mel-manifest/pull/202 to pick up the kmod fix upstream.

PR #559, opened 2015-09-28
Ls1 wic support

skrishnakar:


skrishnakar:

meta-mel/fsl-arm/recipes-bsp/rcw/rcw_git.bbappend
```diff
@@ -0,0 +1,6 @@
+do_deploy_append () {
+	cd ${DEPLOYDIR}
```
kergoth: This will cause problems if another bbappend is applied after this one, as you don't return to the previous directory.

PR #560, opened 2015-09-28
p1020rdb: Add WIC support and wks files

arehman199:


arehman199:
Specified wrong uboot binary, hence closing, will be sending a new PR in a short while.

PR #561, opened 2015-09-28
Use existing, working upstream auto-mount handling

kergoth:


kergoth:
Replacing with another pull req to address various issues with the meta-mel systemd bits.

PR #562, opened 2015-09-28
Use DISTRO_EXTRA_RRECOMMENDS for udev-extraconf

kergoth:
Automount is a distro decision, so it should be done in the distro, not hardcoded in the bbappend in our layer.


kergoth:
Replacing with another pull req to address various issues with the meta-mel systemd bits.

PR #567, opened 2015-09-30
remove redundent wks files

fahadusman:
Drop redundant machine specific .wks files and use generic 2g, 4g, and 8g versions of .wks files based on sdimage-bootpart.wks from poky.


conf/local.conf.sample
```diff
@@ -280,3 +280,14 @@ BB_DISKMON_DIRS = "\
...
+# WIC image type support
+IMAGE_FSTYPES ?= "tar.bz2 wic.bz2"
+
+# dosfstools is needed to create the boot partition, mtools to copy to it
+IMAGE_DEPENDS_wic_append = " dosfstools-native mtools-native"
+
+# Uncomment one of these for specific SD card sizes (default=4g)
+WKS_FILE ?= "sdimage-bootpart-4g.wks"
+# WKS_FILE ?= "sdimage-bootpart-2g.wks"
+# WKS_FILE ?= "sdimage-bootpart-8g.wks"
```
kergoth: I don't think this is appropriate. Using generic wks files is good, but not every MACHINE is guaranteed to work with them. I think this block still belongs in the machine .conf files, or perhaps a .inc which the machines include. For example, I'm pretty sure this is just plain wrong for the qemu machines, which we have to support.

Noor-Ahsan: @kergoth so you want this to be added in append file for each BSP we are doing?

kergoth: I don't see any other way to handle it, unless we want to break qemu support, as well as any BSPs we haven't done the wic support for yet.

Noor-Ahsan: OK. I missed that point. We will rework it accordingly.

skrishnakar:
@fahadusman The 4g as default will create "4GB" of wic file in all cases even If the actual rootfs size is 1GB. Every new rootfs image size will be 4GB by default on disk ? I know we are doing extra-fill by 0.

conf/local.conf.sample
fahadusman: For the machines that require machines-specific changes in wks files, we will use a set of wks files with the same name in a more specific layer but you're right, it won't work with qemu machines.

kergoth:
[**@skrishnakar**](https://github.com/skrishnakar) commented on [Sep 30, 2015, 10:06 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/567#issuecomment-144478639 "2015-09-30T17:06:28Z - Replied by Github Comment Enhancer"):
> @fahadusman The 4g as default will create "4GB" of wic file in all cases even If the actual rootfs size is 1GB. Every new rootfs image size will be 4GB by default on disk ? I know we are doing extra-fill by 0.

For one, we use wic.bz2, not wic, so it's like 80mb, not 4gb, and for two, even using wic, the image is written sparse, so while it looks 4gb, it doesn't actually take up that much space on disk.

conf/local.conf.sample
kergoth: [**@fahadusman**](https://github.com/fahadusman) commented on [Sep 30, 2015, 10:07 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/567/files#r40821213 "2015-09-30T17:07:36Z - Replied by Github Comment Enhancer"):
> For the machines that require machines-specific changes in wks files, we will use a set of wks files with the same name in a more specific layer but you're right, it won't work with qemu machines.

Overridding the wks without changing its name, but substantially changing its content, is not ideal. If it needs its sd card done differently, we'll need new wks files with appropriate names and WKS_FILE set appropriately in the machine .conf.

Noor-Ahsan:
What is the appropriate place for 
# Uncomment one of these for specific SD card sizes (default=4g)
# WKS_FILE ?= "sdimage-bootpart-2g.wks"
# WKS_FILE ?= "sdimage-bootpart-8g.wks"


I think it should go to local.conf.sample. when wic.bz2 will be set in image types then it will be used otherwise not.

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Oct 1, 2015, 4:58 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/567#issuecomment-144706444 "2015-10-01T11:58:07Z - Replied by Github Comment Enhancer"):
> What is the appropriate place for
> # Uncomment one of these for specific SD card sizes (default=4g)
> # WKS_FILE ?= "sdimage-bootpart-2g.wks"
> # WKS_FILE ?= "sdimage-bootpart-8g.wks"
> 
> 
> I think it should go to local.conf.sample. when wic.bz2 will be set in image types then it will be used otherwise not.

I think it has to go into the local.conf.append.<machine> for each machine which supports wic by default, until/if it ends up supported in this form by all bsps. I can't see any other possibility that makes sense, despite the duplication. Does anyone else see a better approach?

fahadusman:
@kergoth Looks like all the bsps other than sabre series will need 'sdimage-bootpart.wks', we can add this section to local.conf.sample and sed it out for sabre series bsps. Otherwise we would have to created local.conf.appends for every single bsp we support.

kergoth:
[**@fahadusman**](https://github.com/fahadusman) commented on [Oct 1, 2015, 9:09 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/567#issuecomment-144774136 "2015-10-01T16:09:22Z - Replied by Github Comment Enhancer"):
> @kergoth Looks like all the bsps other than sabre series will need 'sdimage-bootpart.wks', we can add this section to local.conf.sample and sed it out for sabre series bsps. Otherwise we would have to created local.conf.appends for every single bsp we support.

Don't forget the qemu machines. Other than that, this doesn't seem unreasonable.

Noor-Ahsan:
@kergoth qemu machine is not having wic image type. So having these variable set in local.conf will have no impact on Qemu. So I dont see any issue putting these in local.conf.sample.

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Oct 1, 2015, 11:25 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/567#issuecomment-144808885 "2015-10-01T18:25:07Z - Replied by Github Comment Enhancer"):
> @kergoth qemu machine is not having wic image type. So having these variable set in local.conf will have no impact on Qemu. So I dont see any issue putting these in local.conf.sample.

I'm not sure it's a great idea to have examples in local.conf that don't have any effect. Maybe if we document it in the comments.

Noor-Ahsan:
@Noor-Ahsan<https://github.com/Noor-Ahsan> commented on Oct 1, 2015, 11:25 AM MST<https://github.com/MentorEmbedded/meta-mentor/pull/567#issuecomment-144808885>:

@kergoth<https://github.com/kergoth> qemu machine is not having wic image type. So having these variable set in local.conf will have no impact on Qemu. So I dont see any issue putting these in local.conf.sample.

I'm not sure it's a great idea to have examples in local.conf that don't have any effect. Maybe if we document it in the comments.

Let’s put some extra comments in local.conf.sample to make it clear.

fahadusman:
@kergoth , @Noor-Ahsan please merge the changes till this point if everyone is fine with the changes in this pull request. And we can handle the examples to change the image size, in another pull request.

PR #575, opened 2015-10-06
pull in fsl-networking by default for mel ppc bsps

arehman199:
Pull in meta-fsl-networking for all four PPC BSPs currently
supported by MEL i.e p1010rdb, p1020rdb, p2020rdb and t4240rdb-64b


kergoth:
This seems like a reasonable approach to me if we want it to remain optional, but also want it included by default. If it's mandatory for any of the machines, it should be pulled in via LAYERDEPENDS.

PR #578, opened 2015-10-09
setup-environment: Adding toaster work around fix

sharidas:
This patch creates a work around to run toaster.
The toaster_configuration patch creates temporary git
repos for the layers used by the user in bblayers.conf.
It never forces user to use toaster and it also doesn't
erases the git history for the development environment.

### DIFF
--- a/scripts/setup-mel-builddir
+++ b/scripts/setup-mel-builddir
@@ -37,6 +37,8 @@ Options:
                and the layers
   -l LAYERS    Space-separated list of layer names for additional layers you
                want to be included in the configuration
+  -t           Create toaster configuration file in the build directory with layers
+               configured
   -f           Force overwrite of existing configuration files in the build
                directory. This includes local.conf and bblayers.conf, but not
                site.conf
@@ -143,7 +145,8 @@ process_arguments () {
     OEROOT=
     extralayers=
     force_overwrite=
-    while getopts b:o:l:fh opt; do
+    toasterconfig=
+    while getopts b:o:l:fth opt; do
         case "$opt" in
             b)
                 BUILDDIR="$(abspath $OPTARG)"
@@ -153,6 +156,8 @@ process_arguments () {
                 ;;
             l)
                 extralayers="$OPTARG"
+                ;;
+            t)  toasterconfig="yes"
                 ;;
             f)
                 force_overwrite="-f"
@@ -309,4 +314,8 @@ fi
 MEL_DISTRO="${MEL_DISTRO:-mel}"
 if [ -e $BUILDDIR/conf/local.conf ]; then
     sed -i -e "s/^DISTRO.*/DISTRO = \'$MEL_DISTRO\'/" $BUILDDIR/conf/local.conf
+fi
+
+if [ "$toasterconfig" = "yes" ]; then
+    $layerdir/scripts/toaster_configuration "$BUILDDIR" "$MELDIR"
 fi
--- a/scripts/toaster_configuration
+++ b/scripts/toaster_configuration
@@ -1,0 +1,172 @@
+#!/bin/sh
+#
+# Copyright 2015-2016 Mentor Graphics Corporation
+#
+# This file is licensed under the terms of the GNU General Public License
+# version 2.  This program  is licensed "as is" without any warranty of any
+# kind, whether express or implied.
+
+echo "Creating toaster configuration"
+
+BUILDDIR="$1"
+MELDIR="$2"
+
+
+BBLAYERS=
+configured_layers () {
+   tac "$BUILDDIR/conf/bblayers.conf" | \
+   sed -n -e '/^"/,/^BBLAYERS = /{ /^BBLAYERS =/d; /^"/d; p;}' | \
+   awk {'print $1'}
+}
+
+configured_layers | while read layer; do
+   if [ -d "$layer" ]; then
+       while [ $(dirname $layer) != "${MELDIR}" ];do
+          layer=$(dirname $layer)
+          layername=$(basename $layer)
+       done
+       cd "$layer"
+       if [ ! -d "$PWD/.git" ]; then
+            if [ "$layername" = "poky" ]; then
+                git init . > /dev/null 2>&1
+                git add . > /dev/null 2>&1
+                git commit -m "Initial commit" > /dev/null 2>&1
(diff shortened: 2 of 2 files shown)


sharidas:
@kergoth Kindly review this patch. This patch is a work around to get toaster working with our MEL.

scripts/toaster_configuration
```diff
@@ -0,0 +1,153 @@
...
+   sed -n -e '/^"/,/^BBLAYERS = /{ /^BBLAYERS =/d; /^"/d; p;}' | \
+   awk {'print $1'}
+}
+
+configured_layers | while read layer; do
+   if [ -d "$layer" ]; then
+       #Once testing is done, remove this if.
+       if [ -d "$layer/.git" ]; then
+            rm -fr "$layer/.git"
+       fi
```
kergoth: Am I missing something, or will this just wipe the user's existing git repositories? If this isn't safe to run on setups with existing git repositories, there's no way I'm going to merge it.

kergoth:
Typo in commit message: temperory

scripts/toaster_configuration
```diff
@@ -0,0 +1,153 @@
...
+#
+# Copyright 2015-2016 Mentor Graphics Corporation
+#
+# This file is licensed under the terms of the GNU General Public License
+# version 2.  This program  is licensed "as is" without any warranty of any
+# kind, whether express or implied.
+
+echo "Creating toaster configuration"
+
+BUILDDIR=$*
```
kergoth: This really makes no sense. If you pass multiple arguments, this will just error and freak out on you. Either you accept one argument, or you accept many, and in either case this needs to be quoted. `BUILDDIR="$1"` is appropriate.

scripts/toaster_configuration
```diff
@@ -0,0 +1,153 @@
...
+            rm -fr "$layer/.git"
+       fi
+       cd "$layer"
+       if [ ! -d "$PWD/.git" ]; then
+            git init . > /dev/null 2>&1
+            sleep 1
+            git add . > /dev/null 2>&1
+            sleep 1
+            git commit -m "Initial commit" > /dev/null 2>&1
+            sleep 1
```
kergoth: What's with the pointless sleeps? git doesn't have timing issues, this is pointless.

scripts/toaster_configuration
```diff
@@ -0,0 +1,153 @@
...
+MACHINE=$(grep -i "MACHINE ??=" "$BUILDDIR/conf/local.conf" | awk {'print $3'} | sed -e s:\"::g)
+cat <<EOF > "$BUILDDIR"/toasterconf.json
+{
+    "config": {
+        "MACHINE"    : "$MACHINE",
+        "DISTRO"     : "$DISTRO",
+        "IMAGE_INSTALL_append": "",
+        "PACKAGE_CLASSES": "package_ipk",
+        "EXTERNAL_TOOLCHAIN": "",
+        "MGLS_LICENSE_FILE": "",
```
kergoth: What about LM_LICENSE_FILE?

scripts/toaster_configuration
sharidas: Sorry, I missed that port. Thanks @kergoth  for pointing it out. I forgot to remove that code. Will remove it right away.

scripts/toaster_configuration
sharidas: @kergoth In the toaster configuration where I built console-image, This is what I had 
"EXTERNAL_TOOLCHAIN": "/home/sujith/codebench-linux-install-2014.11-96-arm/codebench",
"MGLS_LICENSE_FILE": "8224@svr-alm-eng-03.alm.mentorg.com",
I didn't added the LM_LICENSE_FILE. Intention of keeping this variable empty is so that user can point to the path in the toaster UI.

sharidas:
@kergoth  I have updated the patch.

scripts/toaster_configuration
```diff
@@ -0,0 +1,146 @@
...
+   if [ -d "$layer" ]; then
+       cd "$layer"
+       if [ ! -d "$PWD/.git" ]; then
+            git init . > /dev/null 2>&1
+            git add . > /dev/null 2>&1
+            git commit -m "Initial commit" > /dev/null 2>&1
+            cat <<EOF >> .git/config
+[remote "mel"]
+        url = git://"$layer".git
+        projectname = meta-fsl-arm.git
```
kergoth: Hardcoded project name to meta-fsl-arm?

sharidas: @kergoth Updated the patch, by removing hardcoded project name.

kergoth:
This will cause us to have local changes in our local poky clones in our development environments, which could be accidentally checked in, and which will block `repo sync` from succeeding. Given we're already stuck with a poky fork for this release, we could always apply them there.

Also, this will run even when we have no desire or intention to run toaster, could we use a wrapper script such that it initializes a config if none exists already whenever you run 'toaster', instead? Hmm.

sharidas:
@kergoth Yes we can do that. That shouldn't be an issue.

scripts/setup-mel-builddir
```diff
@@ -310,3 +315,7 @@ MEL_DISTRO="${MEL_DISTRO:-mel}"
 if [ -e $BUILDDIR/conf/local.conf ]; then
     sed -i -e "s/^DISTRO.*/DISTRO = \'$MEL_DISTRO\'/" $BUILDDIR/conf/local.conf
 fi
+
+if [ "$toasterconfig" = "yes" ]; then
+    $layerdir/scripts/toaster_configuration "$BUILDDIR"
+fi
```
kergoth: This isn't an unreasonable approach. Of course, in the other mechanism, automatically running it, it added overhead to every setup environment execution, but with this one, we'd have to update the docs. I'd suggest running it by kathy and possibly @challinan

sharidas:
@kergoth  Kindly hold this PR. There are few issues which are causing problems with toaster. So will have to wait till they are resolved.

kergoth:
Understood, thanks for the update.

sharidas:
@kergoth I am done with my changes in the script. 
@Noor-Ahsan In the updated script, I try to add EXTERNAL_TOOLCHAIN to the toasterconf.json file. And leave MGLS_LICENSE_FILE empty as of now.

PR #579, opened 2015-10-13
canned-wks: reduce the size of rootfs partition

fahadusman:
Certain SD-Cards have less actual capacity than what they claim, so reduce
the size of rootfs partition so that we may not run out of space.


kergoth:
Looks reasonable to me, though the 2g one could probably be in megs rather than kilobytes like the rest if we're rounding down anyway.

fahadusman:
It might be appropriate to drop the 2g versions, since 2GB memory cards shouldn't be too common these days.

kergoth:
Agreed.

PR #581, opened 2015-10-15
systemd: drop SD/MMC automount rules

abelal:
The automount rules generated a failure on devices where
installation is possible through grub EFI modes and in
such circumstances the rules need to be deleted explicitly.
Additionally, the automounting of block devices is
already handled by the automount.rules from base
systemd recipe.
This has been tested against AMD platforms as well as imx6solox.


abelal:

kergoth:
This is good, I think I might even have this fixed locally somewhere as a part of reviewing all our old systemd modifications. Nicely done.

PR #583, opened 2015-10-15
linux-qoriq: backport patch to fix toc alignment

arehman199:
Backport commit id 5e95235 from mainline kernel to fix hang
caused by TOC misalignment.

Recent toolchains force the TOC to be 256 byte aligned. We need
to enforce this alignment in our linker script, otherwise pointers
to our TOC variables (__toc_start, __prom_init_toc_start) could
be incorrect

JIRA: SB-6017


kergoth:
The patch is missing `Upstream-Status: Backport` above the patch header.

arehman199:
Updated patch as per @kergoth's comment. @Noor-Ahsan kindly review.

Noor-Ahsan:
Is this only for T4240 or applicable on other BSPs too.

arehman199:
@Noor-Ahsan This adds a line in the vmlinux linker script that is common to all powerpc targets.
The added line is in the else clause of a `#ifdef CONFIG_PPC32` statement.
So it will only effect non PPC32 targets which **in our case** is T4240RDB only.
For PPC32 targets it will have no effect.

PR #585, opened 2015-10-20
setup-mel-builddir: set EXTERNAL_TOOLCHAIN here

kergoth:
Rather than relying on it being set in the templates coming out of
archive-release, we can just check to see if the codebench path actually
exists, and if so, adjust it when creating local.conf from local.conf.sample
in our setup scripts.

This fixes EXTERNAL_TOOLCHAIN not being set correctly for the qemu machines.

JIRA: SB-6115


Noor-Ahsan:
nice :)

PR #588, opened 2015-10-25
opkg: apply Paul Barker's cache fixes

kergoth:
His patch series from the opkg-devel list changes the package cache to use
checksums rather than the url for the filename, avoiding the filename length
issue.

JIRA: SB-6120

I tested this locally, and it fixed the problem here.


kergoth:
Paul indicates that these have had limited testing, but we need this fixed, and it does fix the problem here.

kergoth:
Note that the memory leak fix patch from his series was *not* applied, as it wasn't directly related, and applying it causes libc double-free failures.

PR #593, opened 2015-10-27
rrdtool: add recipe from meta-oe, as of ff477aa

arehman199:
Add updated recipe from meta-oe as of commit id
ff477aa4e6199b3433ab315a5533ed89f09ad9c9 to fix a number of
QA warnings(file-rdeps, useless-rpaths, installed-vs-shipped).

JIRA: SB-6141

### DIFF
--- a/meta-mentor-staging/recipes-extended/rrdtool/rrdtool_1.5.4.bb
+++ b/meta-mentor-staging/recipes-extended/rrdtool/rrdtool_1.5.4.bb
@@ -1,0 +1,124 @@
+SUMMARY = "High performance data logging and graphing system for time series data"
+HOMEPAGE = "http://oss.oetiker.ch/rrdtool/"
+
+LICENSE = "GPLv2"
+LIC_FILES_CHKSUM = "file://COPYRIGHT;md5=3349111ed0533471494beec99715bc9d"
+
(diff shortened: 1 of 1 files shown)


kergoth:
Why was this added, and why in meta-mentor-staging? Did we really have to replace the entire recipe just to fix a couple warnings? That seems far from ideal, and will most likely impose maintenance overhead later on.

arehman199:
I added this in staging as it belonged to upstream.

The 2nd option I had was to backport the four commits separately that fixed the warnings.
3rd option was to incorporate all the changes into one commit.

I opted not to..
..go for 2nd option because of the maintenance overhead
..go for the 3rd option as it would have been difficult to reference the four commits in the commit message(as in what changes were being introduced by which commit)

@kergoth Kindly advise, if I need to
a) revert this commit as soon as the layer is updated
b) or send a PR now to revert this commit and only introduce the required changes.

kergoth:
I either either of 2 or 3 would be better than copying recipes wholesale from other layers, but don't worry about it, we'll just remove it when we work on pushing things upstream after the release, and leave it as is in cedar.

arehman199:
@kergoth Thanks for the insight.

PR #600, opened 2015-11-03
libgles-omap3: quiet many warnings

kergoth:


kergoth:
I'm not 100% happy with this way of handling file-rdeps, but I can't find a cleaner way at this time. There aren't many ways to silence those.

PR #608, opened 2015-11-10
mesa-demos: drop obsolete demos

abelal:
mesa-demos using EGL_MESA_screen_surface should be dropped
as the mentioned surface extension has been obsoleted in
mesa and the demos do not work.

Patch imported and backported from
http://cgit.freedesktop.org/mesa/demos/commit/?id=74443c6ee79f3251f643ea05e94df58183e37d0d


abelal:
@Noor-Ahsan, please review.

drewmoseley:
Looks good for MEL configurations to me.
Acked-by: Drew Moseley \<drew_moseley@mentor.com\>

@abelal do we need something conditionally in meta-amd to handle the open builds or do you expect this patch to make it's way to poky at some point?

kergoth:
The patch is in meta-mentor-staging, which means its destined for upstream eventually, most likely shortly after the cedar release.

abelal:
@drewmoseley, as @kergoth pointed out the intention is that this will make its way upstream. I do not think we need conditional handling in meta-amd for this as it is minor.

PR #611, opened 2015-11-10
python: ensure usage of native modules while cross compiling

abelal:
During build of shared modules the interpreter is trying to
import modules from the freshly built target side directories.
This fails where the generated module code is target architecture
specific and the build fails with an illegal instruction error.
The issue can be reproduced for any machine by adding the
-march=bdver4 flag inserted in the TUNE_CC_ARGS which generates
quite target specific code.
We fix this by modifying the PYTHONPATH variable which augments
the default search path.

### DIFF
--- a/meta-mentor-staging/recipes-devtools/python/python/0001-ensure-usage-of-native-modules-while-cross-compiling.patch
+++ b/meta-mentor-staging/recipes-devtools/python/python/0001-ensure-usage-of-native-modules-while-cross-compiling.patch
@@ -1,0 +1,31 @@
+From 4eb30173f2d254f19d5ef18fabed3bdb2e5c1182 Mon Sep 17 00:00:00 2001
+From: Awais Belal <awais_belal@mentor.com>
+Date: Fri, 30 Oct 2015 17:47:07 +0500
+Subject: [PATCH] ensure usage of native modules while cross compiling
+
+The PYTHONPATH variable augments the default search path and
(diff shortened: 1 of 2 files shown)


abelal:

kergoth:
I'm assuming this isn't an issue in most cases as either the binary modules are completely different arch, so it doesn't even attempt to import them in that case? E.g. arm? So it's only when the arch is the same but the tuning is completely wrong, so the import is attempted, but fails?

abelal:
@kergoth, even with arm the imports are done and they succeed probably because the binary modules do not have architecture specific code generated. I checked this by editing the setup.py script and did a 'print re.__file__' right after the first line that imports re. This prints the module location to be one that is from the target python build dir and not sysroot.
With -march=bdver4 the generated code is probably different to an extent that it generates an illegal instruction when loaded.
That said, should the build process use the target side modules or native while importing?

kergoth:
We shouldn't be running the python from python-native and import target modules, we already have them in the native sysroot, so afaik this is a correct fix, I was just wondering why folks have been cross-compiling python with oe/yocto successfully for 8+ years without this :)

abelal:
@kergoth, thanks a lot for taking care of this. Just to clarify that this anomaly came in with python upgrade from 2.7.3 to 2.7.9 where the Makefile was updated rigorously. Also, I confirmed that 2.7.3 was indeed importing and building with the modules from sysroot rather than target.

PR #613, opened 2015-11-11
Switch to use of nss myhostname

kergoth:
Rather than hardcoding our hostname in /etc/hosts to ensure it resolves, we
can use the nss service/library via nsswitch.conf. systemd includes it
already, but needed to enable it, and the nss-myhostname package provides it
for sysvinit images. This is cleaner, more flexible, and avoids the assumption
that the hostname is always the MACHINE, which our netbase bbappend was
making.


kergoth:
Naturally this is intended for after our release.

PR #625, opened 2015-12-02
pulseaudio: include missing bluetooth modules

fahadusman:
We need the pulseaudio-module-bluetooth-discover even in the case of bluez5
because it in turn loads the correct version of bluez-discover module. And
in case of bluez5, installing just pulseaudio-module-bluetooth-discover is
not sufficient, we need pulseaudio-module-bluez4-discover as well in that
case.

Jira: http://jira.alm.mentorg.com:8080/browse/SB-6280


fahadusman:
I haven't verified it with bluez4 but theoretically it should work.

kergoth:
Looks quite reasonable to me.

fahadusman:
Thanks @kergoth.

PR #629, opened 2015-12-08
toaster: Fix layers created in toasterconf.json file

sharidas:
The layer information in toasterconf.json file is in the reverse
order compared to bblayers.conf. Hence toaster creates BBLAYERS
in the reverse order. This was the root cause for SB-6364.

JIRA: SB-6372


kergoth:
This is some terrible, overly complex shell code. Not to mention spaces in the layer paths would break horribly due to how the iteration is done now.

PR #631, opened 2015-12-09
Merge from cedar.

kergoth:


kergoth:
Also revert the distro version lockdown, since master needs to stay floating/snapshot.

PR #637, opened 2015-12-15
MF-309 - Add windows ADE to mel.conf

darknighte:


darknighte:
Hit return to quick.  This should be against the cedar branch.

PR #638, opened 2015-12-15
Add Windows ADE class to mel.conf

darknighte:
Signed-off-by: Sean Hudson <sean_hudson@mentor.com>

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -173,7 +173,7 @@ ADE_SITENAME ?= "ADE for ${ADE_IDENTIFIER}"
 ADE_TOP_CATEGORY ?= "${DISTRO_NAME} ADE"
 
 # If meta-mentor-private is available, pull in the populate-ade class
-ADE_IMAGE_CLASS = "${@'populate_ade' if 'mentor-private' in '${BBFILE_COLLECTIONS}'.split() else ''}"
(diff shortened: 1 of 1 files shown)


darknighte:
@Noor-Ahsan, as we discussed, here is the one line change to remove the need to modify the local.conf to support the windows ADE.  Of course, I don't think we should merge this unless @kergoth and @bikrambhola are comfortable doing so.

kergoth:
Looks fine to me. I think we'll want to consolidate and refactor the classes for dogwood, but this looks fine for now. I'll let @Noor-Ahsan handle the merge, given the risks involved with changes at this point in the release cycle.

Noor-Ahsan:
@bikrambhola Can you review this change and give your suggestion on including this. RIght now we have to include populate_win_ade in IMAGE_CLASSES in local.conf for creating ade. After this change it will not be needed. Will you be able to handle the testing if we add it at this point.

bikrambhola:
@Noor-Ahsan I am Ok with the changes. Since the impact is only on build side and not on functionality, we will run an automated build test to confirm. Please merge. Also need to inform @Dolly and @ulaset for corresponding jenkins build script changes and  change to the Doc section accordingly.

Noor-Ahsan:
@darknighte  kindly coordinate with unika to update the doc. We don't need local.conf change now.

PR #640, opened 2015-12-15
mel: use buildhistory-collect-srcrevs

kergoth:
Rather than dumping/restoring the BB_URI_HEADREVS, use
buildhistory-collect-srcrevs, which emits SRCREV_pn- lines to lock down the
autorev recipes.


kergoth:
This is one less thing to have to maintain ourselves.

PR #641, opened 2015-12-17
mel: fix $layer/downloads PREMIRROR

frederiksdun:
This actually does not break usage of $layer/downloads
	as premirror, but the next premirror because python
	expands the \n, so just escape ht backslash


kergoth:
MIRRORS/PREMIRRORS are actually newline separated, so expansion of \n into an actual newline is harmless. The code just replaces a \\n with \n when processing. See https://github.com/openembedded/bitbake/blob/50d7ffa697764d84d96f26a3ae58a9dacde7b31e/lib/bb/fetch2/__init__.py#L543-L544.

kergoth:
Closing, I replied to this 26 days ago indicating this wasn't needed, but there was no response. If there's a real issue here, re-open. Thanks.

PR #652, opened 2016-02-02
Cedar-async commits

fahadusman:
@Noor-Ahsan , please use this pull request to pull in the changes from cedar-async to cedar branch. We can then remove the cedar-async branch.


meta-mel/conf/local.conf.append.cyclone5
```diff
@@ -0,0 +1,3 @@
+PREFERRED_PROVIDER_virtual/bootloader = "u-boot"
+UBOOT_MACHINE = "socfpga_cyclone5_defconfig"
```
kergoth: Since this snippet gets appended to local.conf, where the user/customer will see it, I'd recommend adding one or more comments to this so they know why it's being done.

PR #674, opened 2016-02-29
mkefidisk.sh: Fix bashism when writing to startup.nsh

sharidas:
Fix bashism when writing to startup.nsh. The back slash
are not taken care on dash. With this change it is taken
care.

JIRA: SB-6725


sharidas:
@Noor-Ahsan @skrishnakar Kindly review my patch.

skrishnakar:
Looks good. Thanks!

kergoth:
This is terribly done. Rather than using portable shell constructs like single quotes or prints instead of echo you hardcode logic to one shell?

kergoth:
printf, that is.

PR #676, opened 2016-03-01
mkefidisk.sh: Replace echo with printf

sharidas:
Remove echo with printf for reusability of the script.
Also remove checks for particular shell. This one line
change should fix the issue.

SB-6725

### DIFF
--- a/meta-mel/intel/scripts/mkefidisk.sh
+++ b/meta-mel/intel/scripts/mkefidisk.sh
@@ -441,13 +441,7 @@ if [ -d $ROOTFS_MNT/etc/udev/ ] ; then
 fi
 
 # Add startup.nsh script for automated boot
-if [ -L /bin/sh ]; then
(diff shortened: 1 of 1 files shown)


sharidas:
As per the comments from @kergoth , I have made the changes into this patch. @Noor-Ahsan  @skrishnakar kindly review the patch.

kergoth:
Thanks for this, looks much better. For reference, http://pubs.opengroup.org/onlinepubs/9699919799/ is an extremely helpful resource when you're wondering whether something is portable or not. This tells you what's posix/sus. I've heard good things about [Beginning Portable Shell Scripting](http://www.amazon.com/Beginning-Portable-Shell-Scripting-Professional/dp/1430210435) (and not just because the author works on yocto and I talk to him on a regular basis), but haven't read it yet myself, I tend to just accrue knowledge from experimentation and the single unix specification..

PR #684, opened 2016-03-04
Adjust for availability of NO_GENERIC_LICENSE upstream

kergoth:
This variable is used in the firmware recipe(s) to indicate that no generic license file is provided.


kergoth:
This quiets the warnings about NO_GENERIC_LICENSE in linux-firmware, as we were providing 3 of its licenses as generic license files in meta-mel-support.

Noor-Ahsan:
@kergoth is this for dogwood?

kergoth:
This particular pull request is, but we might want it for cedar async as well, if we want to quiet those warnings, anyway.

PR #691, opened 2016-03-11
add java support for console and core-image-sato for cyclone5

arehman199:
Conditionally add Java support to console image or
core-image-sato if ENABLE_JAVA is set to "1" in local.conf.

JIRA: SB-6588


Noor-Ahsan:
we will improve the java inclusion python function in next PR. for now its good

PR #694, opened 2016-03-15
mplayer2 support

abelal:
This fixes issues while building/deploying mplayer2 which is now a required package for Cedar (asyncs).


abelal:
@Noor-Ahsan, please review.

PR #695, opened 2016-03-16
Minnow wic image

skrishnakar:


Noor-Ahsan:
@skrishnakar are you planning to push this class upstream. if that is not the case then it should be in mel-support or mel layer not in stagging.

skrishnakar:
@Noor-Ahsan No, The update image class has been split-up with fstype and extensions, this patch will only be applicable to MEL. I will update the PR and re-send it.

skrishnakar:
@Noor-Ahsan Updated the PR and move the baseclass to meta-mel.

PR #697, opened 2016-03-16
weston: fix build failure due to wrong syntax

arehman199:
Use correct conditional syntax for appending to the DEPENDS
variable. Earlier syntax was overriding the items previously
assigned to DEPENDS thus causing build-deps warnings and build
failures resulting from missing dependencies.

libinput, which also pulls in libevdev, is already in DEPENDS
in the main recipe, so no need to add these packages here
again.

JIRA: SB-6825

### DIFF
--- a/meta-mel/meta-altera/recipes-graphics/wayland/weston_1.8.0.bbappend
+++ b/meta-mel/meta-altera/recipes-graphics/wayland/weston_1.8.0.bbappend
@@ -1,4 +1,4 @@
-DEPENDS_cyclone5 += "freerdp libevdev libinput"
+DEPENDS_append_cyclone5 = " freerdp"
 
 # Enable building of rdp-compositor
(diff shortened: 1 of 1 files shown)


kergoth:
Why is this in meta-mel? How is a missing dep not something appropriate for upstream?

Noor-Ahsan:
@kergoth  yeah its problem on our side 

DEPENDS_cyclone5 += "freerdp libevdev libinput"

in meta-mentor/meta-mel/meta-altera/recipes-graphics/wayland/weston_1.8.0.bbappend

I am working with Abdur Rehman on this.

arehman199:
@Noor-Ahsan I've updated the commit as per the discussion we had. Kindly review again.

@skrishnakar Thanks for identifying the redundant inclusion of dependencies. I've fixed those in the updated commit too.

@kergoth upstream recipe builds weston without rdp compositor. We need rdp compositor only for cyclone5 since it does not have a physical display port. freerdp is only required in deps when building weston with rdp compositor enabled, so it seemed like a configuration choice inappropriate for upstream.

PR #703, opened 2016-03-22
Cyclone5 webkit

fahadusman:


kergoth:
At least some of this appears to be general fixes that should go upstream, no? Shouldn't this be split up into both meta-mel and meta-mentor-staging?

fahadusman:
@kergoth yes, the packaging on debug files need to be pushed upstream

fahadusman:
@Noor-Ahsan, @kergoth I have updated the pull request and moved that change to meta-mentor-staging.

meta-mel/meta-altera/recipes-sato/webkit/webkitgtk_2.10.7.bb
```diff
@@ -76,3 +76,5 @@ FILES_${PN} += "${libdir}/webkit2gtk-4.0/injected-bundle/libwebkit2gtkinjectedbu
 
 # http://errors.yoctoproject.org/Errors/Details/20370/
 ARM_INSTRUCTION_SET = "arm"
+
+DEBUG_FLAGS_remove = "-g"
```
kergoth: This failure doesn't occur with non-mel builds?

fahadusman: @kergoth this only occurs when we build webkitgtk with Codebench, and it builds fine with internal toolchain.

PR #710, opened 2016-03-29
intel-corei7-64: bundle wic image in release-artifacts

skrishnakar:
JIRA: SB-6983

Since we have different mechanism employed for WIC
image creation for minnowmax due to processing method
involved in generating live boot image, deploy the
WIC image exclusively in release artifacts so that
we have bootable rootfs in binary folder.


skrishnakar:

PR #711, opened 2016-03-29
cyclone5: add tools to create and modify file-systems

fahadusman:
Although we only need to support mounting of already created file-systems,
we are adding tools to create and modify file-systems for consistency with
other BSPs in Cedar release. This should be removed in Dogwood release.


kergoth:
Looks reasonable to me. Aligns it with our other bsps.

PR #714, opened 2016-03-30
Cyclone5 u boot fixes

fahadusman:
* Preserve the u-boot ELF files in deploy folder.
* Correct the SPL binary file-name in sd-card preparation script.


skrishnakar:
The patch looks reasonable ! Thanks @fahadusman

PR #716, opened 2016-03-31
Add MPV

fahadusman:


meta-mel/meta-altera/recipes-multimedia/mplayer/mpv_0.15.0.bb
```diff
@@ -0,0 +1,66 @@
...
+}
+
+do_compile () {
+    ../waf build
+}
+
+do_install() {
+    ../waf install --destdir=${D}
+}
+
```
abelal: I believe you'll need to override the do_clean as well just like the other functions above. Kindly check.

meta-mel/meta-altera/recipes-multimedia/mplayer/mpv_0.15.0.bb
```diff
@@ -20,8 +20,6 @@ SRC_URI[waf.sha256sum] = "01bf2beab2106d1558800c8709bc2c8e496d3da4a2ca343fe091f2
 
 inherit pkgconfig
 
-# Note: both lua and libass are required to get on-screen-display (controls)
-PACKAGECONFIG ??= "lua libass"
```
abelal: Rather than doing it this way, why not do a PACKAGECONFIG_remove_cyclone5 through a bbappend perhaps just to make sure we can use this stuff on other BSPs with controls enabled if needed.

abelal:
Shouldn't these be moved to a more generic location rather than meta-mentor/meta-altera so it is usable for other BSPs when needed? The core-image-sato append can reside in meta-altera though.

meta-mel/meta-altera/recipes-multimedia/mplayer/mpv_0.15.0.bb
kergoth: [**@abelal**](https://github.com/abelal) commented on [Mar 31, 2016, 7:23 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/716/files#r58062020 "2016-03-31T14:23:55Z - Replied by Github Comment Enhancer"):
> I believe you'll need to override the do_clean as well just like the other functions above. Kindly check.

do_clean wipes all of the separate build directory, it doesn't run make clean. As long as the recipe isn't inheriting autotools-brokensep, anyway, which this one does not. So clean will work just fine without being overridden.

kergoth:
[**@abelal**](https://github.com/abelal) commented on [Mar 31, 2016, 7:28 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/716#issuecomment-203964587 "2016-03-31T14:28:24Z - Replied by Github Comment Enhancer"):
> Shouldn't these be moved to a more generic location rather than meta-mentor/meta-altera so it is usable for other BSPs when needed? The core-image-sato append can reside in meta-altera though.

I'd agree with this, the ffmpeg and mpv recipes shouldn't be in meta-altera.

fahadusman:
@Noor-Ahsan I have updated the pull request, moved the recipes to meta-mentor-staging now.

meta-mel/meta-altera/recipes-sato/images/core-image-sato.bbappend
```diff
@@ -3,6 +3,7 @@ IMAGE_INSTALL_append_cyclone5 = " xserver-xorg-xvfb \
 				  mesa-megadriver \
 				  xserver-xorg-extension-glx \
 				  webkitgtk \
+				  mpv \
```
Noor-Ahsan: @fahadusman  We just have to test it. We don't have to include it :).

meta-mentor-staging/recipes-multimedia/mplayer/mpv_0.15.0.bb
```diff
@@ -22,6 +22,8 @@ inherit pkgconfig
 
 # Note: both lua and libass are required to get on-screen-display (controls)
 PACKAGECONFIG ??= "lua libass"
+PACKAGECONFIG_remove_cyclone5 = "lua libass"
+
```
Noor-Ahsan: Can we do this in meta-altera folder in bbappend

abelal:
@fahadusman, this looks pretty good now but please have a look at https://github.com/drewmoseley/meta-amd/tree/4k-demo specifically the commit https://github.com/drewmoseley/meta-amd/commit/e3d74465a4d9593e683d7b9039f78aa782e43726. @drewmoseley, had to add some additional dependencies to get it to build correctly just confirm that is covered.

drewmoseley:
@abelal some of those dependencies may be AMD-specific.  At least the VDPAU one.  Once this PR is completed we'll need to make sure we have all the AMD-specific stuff squared away inside meta-amd.

PR #721, opened 2016-04-04
Cyclone-5 MPV

fahadusman:
I have addressed a couple of build time warnings in this pull request.

### DIFF
--- a/meta-mel/conf/local.conf.append.cyclone5
+++ b/meta-mel/conf/local.conf.append.cyclone5
@@ -43,3 +43,6 @@ PREFERRED_PROVIDER_openjdk-7-jre = "openjdk-7-jre"
 
 # Use hard-float support
 DEFAULTTUNE = "cortexa9hf-neon"
+
+# Prefer ffmpeg instead of libav for ffmpeg.
+PREFERRED_PROVIDER_ffmpeg = "ffmpeg"
--- a/meta-mentor-staging/recipes-multimedia/mplayer/mpv_0.15.0.bb
+++ b/meta-mentor-staging/recipes-multimedia/mplayer/mpv_0.15.0.bb
@@ -4,6 +4,7 @@ HOMEPAGE = "http://www.mpv.io/"
 DEPENDS = "zlib ffmpeg jpeg virtual/libx11 xsp libxv \
            libxscrnsaver libv4l libxinerama \
            libxkbcommon \
+           pulseaudio virtual/mesa libdrm \
 "
 
 REQUIRED_DISTRO_FEATURES = "x11"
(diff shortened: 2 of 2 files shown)


fahadusman:
@abelal , @Noor-Ahsan kindly review this pull request.

meta-mel/conf/local.conf.append.cyclone5
```diff
@@ -43,3 +43,7 @@ PREFERRED_PROVIDER_openjdk-7-jre = "openjdk-7-jre"
 
 # Use hard-float support
 DEFAULTTUNE = "cortexa9hf-neon"
+# Use ffmpeg instead of libav
+
+# Prefer ffmpeg instead of libav for ffmpeg.
```
abelal: I believe one of the above comments is just a remnant of changes.

meta-mentor-staging/recipes-multimedia/mplayer/mpv_0.15.0.bb
```diff
@@ -5,6 +5,7 @@ DEPENDS = "zlib ffmpeg jpeg virtual/libx11 xsp libxv \
            libxscrnsaver libv4l libxinerama \
            libxkbcommon \
 "
+RDEPENDS_${PN} += "libpulse libgbm libdrm"
```
abelal: libpulse isn't a build dependency? How are these library references being put in the app if they are not used during build time.

fahadusman:
@abelal I have removed the extra comment from local.conf.append.cyclone5 and handled dependency via DEPENDS instead of RDEPENDS as the runtime library dependencies would be handled automatically.

meta-mentor-staging/recipes-multimedia/mplayer/mpv_0.15.0.bb
```diff
@@ -4,6 +4,7 @@ HOMEPAGE = "http://www.mpv.io/"
 DEPENDS = "zlib ffmpeg jpeg virtual/libx11 xsp libxv \
            libxscrnsaver libv4l libxinerama \
            libxkbcommon \
+           pulseaudio mesa \
```
abelal: wouldn't virtual/mesa be more appropriate?

fahadusman: yes, or maybe virtual/libgl would be more appropriate, since MPV actually needs libgl that mesa provides.

abelal:
Looks good now.
Acked-by: Awais Belal (@abelal)

@Noor-Ahsan, we can merge this now I guess.

fahadusman:
Thanks @abelal , @Noor-Ahsan .

PR #723, opened 2016-04-05
setup-mel-builddir: recurse an extra level in meta* dirs

kergoth:
meta-haswell-wc is an extra level deep in meta-intel, so we need to check one
level deeper for layers. If we recurse into `*/*/*` rather than `meta*/*/*`, the
layer search can take quite a long time due to searching into downloads/ and
cached-binaries/, unless we add exclusions.


kergoth:
I'm not 100% happy with hardcoding the 'meta' prefix assumption, but hardcoding exclusion paths like downloads and cached-binaries isn't ideal either, so I think this is a step in the right direction.

PR #726, opened 2016-04-06
cyclone5: restructure the logic to include commercial multimedia packages

fahadusman:


kergoth:
Looks good to me, thanks for taking the time. I think it's a little more readable this way.

In case you're curious, the reason I suggested use of .= rather than += is so when the boolean isn't enabled, there's no change to the value at all, not even a space. It's a habit I've developed out of awareness of the variable checksumming and sstate -- adding a space to a variable that's included in a checksum will change it, since bitbake isn't aware of variable types yet :)

fahadusman:
Thanks for the guidance @kergoth . I've learned a few new things.

PR #736, opened 2016-04-13
mk-cyclone5-boot-part.sh: modify the way we check root access

fahadusman:
Checking the UID environment variable doesn't work with dash, so use the
'whoami' command.


kergoth:
FYI, 'whoami' isn't mandated to exist by posix. I'm certain it's available in all our supported distros, so I won't object to this, but for future reference using `id -u` is more portable. See http://pubs.opengroup.org/onlinepubs/9699919799/utilities/id.html.

fahadusman:
Thanks @kergoth, I will update the script to use 'id -u'

PR #748, opened 2016-04-21
alsa-lib: Remove this recipe

sharidas:
Remove this recipe as none of the patches are taken.
The entire bbappend is commented. So it doesn't make
any sense to keep this recipe.

JIRA: SB-7150


sharidas:
@Noor-Ahsan , @skrishnakar  kindly review my changes.

Noor-Ahsan:
@sharidas kindly coordinate @kergoth  on this. in Cedar release those were not commented. I think @kergoth  commented patches when he upgraded the version. We can remove it without knowing why we added those patches and do we need them now or not.

kergoth:
Yes, I think this was a temporary measure, as @Noor-Ahsan states, we need to determine which patches we need. See  https://gist.github.com/kergoth/fc5355c87d9f7f6aae6c, which states: "recipes-multimedia/alsa: there are quite a few patches floating around here, likely best to have someone more familiar with these changes weigh in"

kergoth:
Er, wrong button, didn't intend to close, only to comment.

kergoth:
Actually closing, as we don't want to take this approach.

PR #753, opened 2016-04-22
cyclone sato fix

arehman199:


kergoth:
Why is the xserver-nodm-init version change specific to cyclone V? What exactly does this change buy us, and why do we not need it for our other BSPs?

PR #755, opened 2016-04-26
lttng-tools: remove the override

fahadusman:
0001-Fix-regression-tests.patch is not architecture specific so remove
the x86-64 override.

JIRA: http://jira.alm.mentorg.com:8080/browse/SB-7174


skrishnakar:
The patch looks reasonable to me.

PR #769, opened 2016-05-06
Migration to meta-mentor-bsp

kergoth:


kergoth:
https://github.com/MentorEmbedded/meta-mentor-bsp shows what the new repository will look like. It'll be sitting at the root of the repo setup like all our other repositories. Each individual sub-layer will be archived separately, i.e. `meta-mentor-bsp_common_0.tar.bz2`, `meta-mentor-bsp_intel-corei7-64_0.tar.bz2`, and will unpack into the same path where they lived in the internal setup. That is, they'll still appear under meta-mentor-bsp/ in the user's setup, rather than being at toplevel the way the update layers are.

kergoth:
The layers in meta-mentor-bsp will likely also contain our probe configs (currently in meta-mentor-private) and the xml metadata used for the web site (currently in docs).

PR #770, opened 2016-05-11
Update for the current shallow implementation

kergoth:


kergoth:
This needs to be merged at the same time as https://github.com/MentorEmbedded/poky/pull/71.

Noor-Ahsan:
@kergoth this is not merging automatically .... Can you update this?

kergoth:
Ah,  yes, this was also adjusting the bits that are now in meta-mentor-bsp. Will do, and will submit another pull req to that layer.

kergoth:
This has been updated.

PR #783, opened 2016-05-24
Drop some bits which are already upstream

kergoth:


kergoth:
Merging this now, as at least some of this fixes issues with the layer update.

PR #785, opened 2016-05-24
Use LAYERDIR_RE

kergoth:
LAYERDIR_RE, a newly introduced bitbake variable, is a regex-escaped version
of LAYERDIR.


kergoth:
The default value is added for compatibility with bitbake versions lacking the new variable.

PR #791, opened 2016-05-25
mel: add support for .wks.in templates

kergoth:
This will hopefully be merged upstream soon and we can drop it from here, but
in the meantime we can start making use of it.

These files are treated as the contents of a bitbake variable, so usual
bitbake variable references are supported. I considered using another
templating mechanism, for example the one used by yocto-layer, but then we'd
end up largely mapping metadata variables to template fields anyway, which is
a pointless indirection. Let bitbake expand the variables directly instead.

This feature lets us, for example, reference ${APPEND} in --append, and avoid
hardcoding the serial console tty in the wks file, and let the user's changes
to APPEND affect wic the way they do the other image construction mechanisms.
The template is read in and set in a variable at parse time, so changes to the
variables referenced by the template will result in rebuilding the image.

### DIFF
--- a/meta-mel/classes/wic_wks_template.bbclass
+++ b/meta-mel/classes/wic_wks_template.bbclass
@@ -1,0 +1,54 @@
+inherit image_types
+
+python do_write_wks_template () {
+    """Write out expanded template contents to WKS_FULL_PATH."""
+    import re
+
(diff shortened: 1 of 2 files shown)


kergoth:
I switched it from prefuncs to a separate task for clarity and to avoid the need for the oe-core/poky change.

PR #792, opened 2016-05-25
qtmultimedia: drop the gstreamer1.0-plugins-bad dep

kergoth:
qtmultimedia needs gstreamer, not a random collection of plugins with
questionable quality

### DIFF
--- a/meta-mentor-staging/qt5-layer/recipes-qt/qt5/qtmultimedia_%.bbappend
+++ b/meta-mentor-staging/qt5-layer/recipes-qt/qt5/qtmultimedia_%.bbappend
@@ -1,0 +1,3 @@
+# qtmultimedia needs gstreamer, not a random collection of plugins with
+# questionable quality
+DEPENDS_remove_pn-qtmultimedia = "gstreamer1.0-plugins-bad"


Noor-Ahsan:
@kergoth doesn'r plugins-bad have licensing issue. Can't we just make use of plugins-good? Secondly this change should go in meta-qt5-mel layer if want this change.

kergoth:
I think you misread the commit and got it backwards. This commit removes the dep on plugins-bad, it doesn't add it. "Can't we just make use of plugins-good?" -- this is exactly what this commit does, it leaves the dependency on plugins-good and removes the unneeded dependency on plugins-bad.

If it had licensing issues, it wouldn't matter, because again, this commit removes it.

But no, plugins-bad has no licensing issues, it's plugins-ugly that has licensing issues. plugins-bad is just plugins that aren't up to par. See https://gstreamer.freedesktop.org/modules/gst-plugins-bad.html: "GStreamer Bad Plug-ins is a set of plug-ins that aren't up to par compared to the rest. They might be close to being good quality, but they're missing something - be it a good code review, some documentation, a set of tests, a real live maintainer, or some actual wide use."

kergoth:
Regarding meta-qt5-mel, no, I think it's best off in meta-mentor-staging, because I think this belongs upstream in meta-qt5, not stuck in the limbo that is the qt5-mel layer.

Noor-Ahsan:
Sorry I missread it. @sharidas  Kindly review this ..... will it has any impact on things that we have tested?

kergoth:
JaMa pointed me to the reasoning behind this, apparently qtmultimedia has a floating dep on the gstreamer photography interface. With no way to disable it, we have to include plugins-bad to make the build deterministic:
```
commit 0732449ee42bb3582e5c141fa2363e18beb43a9f
Author: Otavio Salvador <otavio@ossystems.com.br>
Date:   Sat Feb 7 18:33:26 2015 -0200

   qtmultimedia: Fix floating dependency gst-plugins-bad

   When using GStreamer 0.10 support, QMake detects if GStreamer
   Photography interface is available. To avoid a floating dependency we
   add the gst-plugins-bad as dependency so it is always present.

   Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>
```

So, closing this out.

kergoth:
I might take a quick look to see what it'd take to add a config option to disable/enable the photography interface support.

PR #794, opened 2016-05-31
initial support for Imx7 based sabresd board

adnan-ali1:
meta-fsl-arm "jethro" provides imx7dsabresd machine file.
. directfb is removed in meta-fsl-arm stating that the project is dead upstream
. set preferred provider to uboot-imx instead of uboot-fslc
. systemd workaround for the random bug when boot hangs at serial-getty service.

Signed-off-by:  Adnan Ali <Adnan_Ali@mentor.com>

### DIFF
--- a/meta-mel/conf/local.conf.append.imx7dsabresd
+++ b/meta-mel/conf/local.conf.append.imx7dsabresd
@@ -1,0 +1,5 @@
+# By default meta-fsl-arm selects u-boot-fslc.
+# Override it and select u-boot-imx as our PREFERED_PROVIDER.
+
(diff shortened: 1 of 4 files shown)


adnan-ali1:
@Noor-Ahsan @skrishnakar  please review.

skrishnakar:
@adnan-ali1 The MEL BSP which does not have own layer and just uses upstream changes then it is recommended to use new layer "meta-mentor-bsp". See other machine changes for reference and send PR against that layer.

Noor-Ahsan:
@skrishnakar  that is for dogwood. this BSP is for Cedar.

PR #797, opened 2016-05-31
mpv: use the waf class

kergoth:


kergoth:
The waf class expects to run ${S}/waf, hence the symlink

kergoth:
I'll re-submit this against meta-oe once libass+mpv is merged upstream.

PR #800, opened 2016-06-02
mel: add vmlinux to KERNEL_IMAGETYPES by default

kergoth:
This ensures vmlinux ends up in tmp/deploy/images/${MACHINE}, next to the
kernel and disk images, which is useful for debugging.


Noor-Ahsan:
I like this.

PR #816, opened 2016-06-17
openssh: use mel override

kergoth:
Changes in the distro layer need to be conditional upon the distro.


kergoth:
This is the only issue I found in meta-mentor and meta-sourcery when doing leaky layer tests.

PR #821, opened 2016-06-28
scripts: Add a script for flashing an SD card with a wic image

wfarnsworth:
This adds the script mksd-wic which writes a wic image to an SD card.
Features include:

* Support for /dev/sd* and /dev/mmcblk* style devices
* Detection of host system partitions ("/", "/home", etc)
* An option to list details of block devices (size of partition, where mounted)
* An option to extend the rootfs to consume free space on the card.

Signed-off-by: Wade Farnsworth <wade_farnsworth@mentor.com>

### DIFF
--- a/scripts/mksd-wic
+++ b/scripts/mksd-wic
@@ -1,0 +1,207 @@
+#!/bin/bash
+#
+# Copyright 2016 Mentor Graphics Corporation
+#
+# This file is licensed under the terms of the GNU General Public License
+# version 2.  This program  is licensed "as is" without any warranty of any
+# kind, whether express or implied.
+
+# List of critical mountpoints.  Any device with a partition
+# mounted to one of these will not be overwritten.
+MOUNTPOINT_BLACKLIST=" \
+    / \
+    /home \
+    /opt \
+    /var \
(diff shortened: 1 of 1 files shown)


wfarnsworth:
@Noor-Ahsan, @darknighte: This is something I've been playing with and I think it's a good fit for Dogwood, as it attempts to prevent a user from overwriting the host rootfs due to, for example, a type in the dd command.  Please review and merge if appropriate.

kergoth:
Seems handy, though it's not at all wic-specific, afaict. I'm also curious why you didn't enhance the existing ddimage script in oe-core (guessing you probably didn't know it existed). The rootfs resizing is interesting, we recently added first-boot-time rootfs resizing to minnowmax. I'd also suggest adding the ability to optionally call out to bmaptool rather than dd, to write a sparse image from a .wic.bz2+.wic.bmap.

Noor-Ahsan:
Does everybody agree to merge this?

wfarnsworth:
At this point I don't see much value of merging this script, since its primary functionality duplicates ddimage.  I will look at merging some of the additional features into it.

PR #822, opened 2016-07-07
Use plymouth with a mel splash screen

kergoth:
Also set a sane default for the APPEND variable while we're at it, though not
all BSPs will make use of it.


kergoth:
The plymouth theme is a modified version of the 'script' default theme, adjusted per some online tutorials to handle differing aspect ratios for the splash screen and whatnot.

kergoth:
@fahadarslan, does this look reasonable? Obviously it needs to be tested on every BSP to make sure it's starting early enough, or switch to an initramfs where appropriate, and kernel cmdlines adjusted for each, since not all obey APPEND.

meta-mel-support/recipes-core/plymouth/plymouth-mel/mel.script
```diff
@@ -0,0 +1,162 @@
...
+     scale_factor =  Window.GetHeight() / bg_image.GetHeight();
+  }
+else
+  {  # Screen ratio is wider than image ratio, we will match the screen width
+     scale_factor =  Window.GetWidth() / bg_image.GetWidth();
+  }
+#scale_factor = Window.GetWidth() / bg_image.GetWidth();
+
+scaled_bg_image = bg_image.Scale(bg_image.GetWidth()  * scale_factor,
+                                 bg_image.GetHeight() * scale_factor);
```
fahadarslan: For line# 3-17, how about following approach to scale image according to screen dimensions

`+width_scale_factor = Window.GetWidth() / bg_image.GetWidth();`
`+height_scale_factor = Window.GetHeight() / bg_image.GetHeight();`
`+`
`+scaled_bg_image = bg_image.Scale(bg_image.GetWidth()  * width_scale_factor,`
`+                                 bg_image.GetHeight() * height_scale_factor);`

fahadarslan:
>Obviously it needs to be tested on every BSP...
@kergoth, right. Probably by adding a verification task for this for each target as part of BSP enablement.

meta-mel-support/recipes-core/plymouth/plymouth-mel/mel.script
kergoth: That'll distort the image when the aspect ratio of the image isn't the same as the aspect ratio of the display, since the width and height factors wouldn't be the same in that case.

fahadarslan: Right, I didn't consider that. 

Minor changes are needed: 
When if condition is true, screen ratio is wider than image ratio so comment on line# 7 needs to be updated. 
Similarly comment in else condition needs to be updated (line# 11, taller instead of wider). 
Comment on line# 14 may be removed. 

Rest looks great. Thanks!

kergoth:
@Noor-Ahsan, I think we should be good to merge this now, then we'll want to add verification tasks for this to each forthcoming BSP, and create another issue to cover the existing ones.

PR #826, opened 2016-07-15
mel: add sourcery tc version to ade env script

kergoth:


kergoth:
Awaiting a confirmation via email that this is sufficient, but we probably want this regardless, they just might want more in addition to this, which we could do as another pull request.

PR #827, opened 2016-08-01
mplayer2: drop changes from our layer

abelal:
For MEL on the required BSPs we now use mpv which is
the better suited option so the changes for mplayer2
are not needed anymore and are irrelevant.


abelal:
@Noor-Ahsan, please review.

PR #830, opened 2016-08-09
archive-release: limit version to the sub-repo

kergoth:
For git repositories, which use commit counts for versions, we want to limit
the commit count to the current path, so it doesn't increase due to changes in
unrelated layers.

JIRA: SB-7773


Noor-Ahsan:
@kergoth  You can now resolve the issue

PR #832, opened 2016-08-12
ADE sysroot path fix

fahadusman:


meta-mentor-staging/classes/sdk_auto_relocate.bbclass
```diff
@@ -17,23 +17,31 @@ SDK_POST_INSTALL_COMMAND_append = 'echo "${env_setup_script%/*}" >"${env_setup_s
...
     if [ -n "\$BASH_SOURCE" ] || [ -n "\$ZSH_NAME" ]; then
         if [ -n "\$BASH_SOURCE" ]; then
             scriptdir="\$(cd "\$(dirname "\$BASH_SOURCE")" && pwd)"
         elif [ -n "\$ZSH_NAME" ]; then
             scriptdir="\$(cd "\$(dirname "\$0")" && pwd)"
         fi
-        env -i "\$scriptdir/${sdkpath_to_bindir}/sdk-auto-relocate" && SDK_RELOCATING=1 . "\$scriptdir/environment-setup-${REAL_MULTIMACH_TARGET_SYS}"
-        unset scriptdir
+
+        if [ -e "\$scriptdir/sysroots/i686-melsdk-linux/usr/bin/sdk-auto-relocate" ]; then
```
kergoth: This hardcodes the sysroot path when the path one line below it does not. Please copy/paste the correct version that doesn't hardcode i686-melsdk-linux :)

meta-mentor-staging/classes/sdk_auto_relocate.bbclass
```diff
@@ -17,23 +17,31 @@ SDK_POST_INSTALL_COMMAND_append = 'echo "${env_setup_script%/*}" >"${env_setup_s
...
+            env -i "\$scriptdir/${sdkpath_to_bindir}/sdk-auto-relocate" && SDK_RELOCATING=1 . "\$scriptdir/environment-setup-${REAL_MULTIMACH_TARGET_SYS}"
+        else
+            echo >&2 "Warning: Unable to find sdk-auto-relocate script"
```
kergoth: Have you tested the script with sdk-auto-relocate installed and without it installed to check sanity in both cases?

fahadusman: I checked it with manually deleting the sdk-auto-relocate.

fahadusman:
@kergoth I have updated the commits.

Thanks.

meta-mentor-staging/classes/sdk_auto_relocate.bbclass
```diff
@@ -17,23 +17,31 @@ SDK_POST_INSTALL_COMMAND_append = 'echo "${env_setup_script%/*}" >"${env_setup_s
...
-        unset scriptdir
+
+        if [ -e "\$scriptdir/${sdkpath_to_bindir}/sdk-auto-relocate" ]; then
+            env -i "\$scriptdir/${sdkpath_to_bindir}/sdk-auto-relocate" && SDK_RELOCATING=1 . "\$scriptdir/environment-setup-${REAL_MULTIMACH_TARGET_SYS}"
+        else
+            echo >&2 "Warning: Unable to find sdk-auto-relocate script"
+        fi
     else
         echo >&2 "Warning: Unable to determine SDK install path from environment setup script location. Please run <installdir>/${sdkpath_to_bindir}/sdk-auto-relocate manually."
+        scriptdir="\$(pwd)"
```
kergoth: I think we might be better off just not doing the sed if we don't know the correct scriptdir, rather than setting it to $PWD and then sed'ing SDKPATH to that. Thoughts?

fahadusman: Sorry, i didn't get your point. At the time we are replacing SDKPATH with $scriptdir during the build, we don't know if we will get the correct value of scriptdir when the script is executed.

fahadusman:
@kergoth I have dropped the commit to replace SDKPATH with $scriptdir. Now the environment-setup script would have to be auto-relocated before it can be used with ADE. The sdk-auto-relocate will still fail for not finding the native lib folder but it correctly relocates the environment-setup script before it terminates. I will add a check to handle absence of native libs in a different PR.

kergoth:
I don't understand. If you drop the sed, it defeats the entire purpose of this pull request. The whole point of this pull request was to let the script go by its own location rather than requiring the relocation script to fix itself. All I suggested was changing the default behavior when the script can't find its own location from $PWD to ${SDKPATH}..

fahadusman:
Oh, I misinterpreted your comment, now I have added that commit back and defaulted scriptdir to ${SDKPATH}, when the script location is not found from the environment.

kergoth:
Looks good, thanks.

PR #836, opened 2016-08-19
plymouth-mel: correctly handle license-checksum error

abelal:
The license-checksum checks are run through populate_lic_qa_checksum()
which does not handle INSANE_SKIP. We drop the checking altogether
by removing it from the ERROR_QA list.

JIRA Ticket: SB-7753


abelal:
@kergoth, please review.
This is as per the discussion on IRC and https://patchwork.openembedded.org/patch/129529/.

kergoth:
It's debatable if this is the right answer, but it'll do for now. Thanks.

PR #839, opened 2016-08-29
toaster_configuration: update this script to generate custom fixtures

sharidas:
As per the latest changes in the upstream django fixutres are used
to populate the default values for toaster instead of fixtures.
This script has been modified for the same. A new custom.xml
file will be generated which will be read by toaster and pre populate
the default values for our local changes.

JIRA: SB-7649


sharidas:
@Noor-Ahsan @skrishnakar Kindly review my changes.

PR #840, opened 2016-08-29
pasplash: Remove psplash from the layer

sharidas:
We are no more using psplash. So we can safely remove
this recipe from this layer.

JIRA: SB-7198


sharidas:

PR #849, opened 2016-09-09
mpeg2dec: Recipe can be removed safely

sharidas:
This recipe can be removed safely. With upstream
meta-altera, tried to build it. But that didn't
showed up any warning.


sharidas:

PR #854, opened 2016-09-16
archive-release: add PDK_DISTRO_VERSION to local.conf.sample if set

kergoth:


kergoth:
It has no effect when unset, so no real risk to merging it.

PR #862, opened 2016-09-28
Revert "pasplash: Remove psplash from the layer"

abelal:
The MEL distro configuration (mel.conf) allows specifying
the SPLASH variable to select between different packages
that provide the mechanism for displaying a splash. psplash
is one of these packages so we shouldn't remove this.

This reverts commit 413cbf8.


abelal:
@Noor-Ahsan, please review.

PR #863, opened 2016-09-29
Add bits to handle the new codebench layout

kergoth:
If $MELDIR/../../toolchains exists, set CODEBENCH_PATH rather than
EXTERNAL_TOOLCHAIN.

JIRA: SB-7881

### DIFF
--- a/scripts/setup-mel-builddir
+++ b/scripts/setup-mel-builddir
@@ -267,7 +267,11 @@ setup_builddir () {
         echo "created for you with some default values."
 
         if [ -d "$MELDIR/../../codebench" ]; then
-            sed -i -e 's,^#\?EXTERNAL_TOOLCHAIN.*,EXTERNAL_TOOLCHAIN ?= "${MELDIR}/../../codebench",' "$BUILDDIR/conf/local.conf"
+            if [ -d "$MELDIR/../../toolchains" ]; then
+                sed -i -e 's,^#\?EXTERNAL_TOOLCHAIN.*,CODEBENCH_PATH ?= "${MELDIR}/../../codebench",' "$BUILDDIR/conf/local.conf"
+            else
+                sed -i -e 's,^#\?EXTERNAL_TOOLCHAIN.*,EXTERNAL_TOOLCHAIN ?= "${MELDIR}/../../codebench",' "$BUILDDIR/conf/local.conf"
+            fi
         fi
 
         (


Noor-Ahsan:
@kergoth can you double check that you are doing the same what you have mentioned in the comment. I think you are doing opposite.

kergoth:
That's not actually the case. It'll make a lot more sense if you review the commits separately, as they do two very different things, but here it is:

We shouldn't change the meaning of EXTERNAL_TOOLCHAIN, as the logic for that is sound, it's just the layout of CodeBench and the particular path to the toolchain which has changed. I don't want the user to have to explicitly set the full path to the toolchain in the new layout, as it includes the target triplet/sys and the CodeBench version.

If we were to keep only one variable, we'd have to see the user-set EXTERNAL_TOOLCHAIN to the codebench path and change its value behind the scenes so the rest of meta-sourcery can do its job, but if we do that, we're modifying a user input variable without their knowledge, and bitbake -e won't reflect what they set it to, which isn't going to work for usability reasons. We could add a warning that we're doing so, but then they'd have no way to quiet the warning, or to prevent the value changing under them. So, it's clear we need a second variable, and this is what the first commit does, introduces a new variable CODEBENCH_PATH, which if set, is used to set EXTERNAL_TOOLCHAIN. It's only used if EXTERNAL_TOOLCHAIN is not set, for compatibility reasons -- we assume if the user set EXTERNAL_TOOLCHAIN explicitly rather than CODEBENCH_PATH, that they know exactly what toolchain they want.

This is straightforward enough, but there's a compatibility issue here with the user's semantics. Previous users expect to be able to set EXTERNAL_TOOLCHAIN and call it done. Setting that to the root of a new CodeBench install is a configuration issue, but we know what the correct values are. Given that, we can fix them, and warn the user we're doing so, so they can correct them and thereby quiet the warnings in the way we couldn't have without adding CODEBENCH_PATH.

I realize it seems like we're adding yet another layer of indirection, but I think there's sense there given what we're working with. If a user sets EXTERNAL_TOOLCHAIN to the root of a new CodeBench install, or to the codebench path inside that install, it will warn them and fix it so CODEBENCH_PATH is set instead, so they know how to fix their local.conf, and continue the build, letting the logic in the first commit set EXTERNAL_TOOLCHAIN and EXTERNAL_TARGET_SYS from that.

I'm open to ideas on how to improve this, but I don't really see how to do so without causing other issues.

kergoth:
Fixing, thanks.

kergoth:
Fixed, thanks.

PR #864, opened 2016-10-05
Revert "perf: Remove this change safely"

sharidas:
We need to revert this patch. The local builds
for minnow max is still failing. There is an
upstream patch posted:
http://lists.openembedded.org/pipermail/openembedded-core/2016-October/127281.html
Once this get merged, we can safely remove our
changes for perf.

This reverts commit 31049561fdca150ece2e18856a94f05e8725b3fc.


sharidas:

PR #866, opened 2016-10-06
Automate setup of the virtualenv for toaster

kergoth:
Depends upon https://github.com/MentorEmbedded/meta-mentor/pull/865, to avoid conflicts between the two pull requests. Only the last 3 commits are unique to this branch.


kergoth:
Updated to add detail to the commit messages.

kergoth:
@Noor-Ahsan , updated to include the update to setup-rh, which is much simper since python-virtualenv can be installed with yum. Tested setup of the venv and start of toaster in a centos7 chroot.

sharidas:
@kergoth I have verified the patchset. It looks ok to me.

PR #869, opened 2016-10-13
psplash: enable systemd support and cleanup

skrishnakar:
JIRA: SB-8147

This patch adds following support:
- Systemd support
- Removes progress bar when built for systemd
- Modified startup service to avoid late init during bootup
- Move from staging to meta-mel


kergoth:
Is this truly mel-specific? It seems like a general improvement to me.

PR #870, opened 2016-10-13
toaster_configuration: Fix the meta layer addition to toaster

sharidas:
This patch fixes the meta layer addition to toaster. Without
this patch the meta layer was added but was not visible in the
UI. With this change it gets fixed.


sharidas:
@Noor-Ahsan @skrishnakar Kindly review my changes.

PR #871, opened 2016-10-13
iperf: Upgrade to iperf3

drewmoseley:
The older, obsolete iperf v2 fails to build with GCCv6.  Upgrade
to v3 to resolve this build failure.

Signed-off-by: Drew Moseley <drew_moseley@mentor.com>

### DIFF
--- a/meta-mel-support/recipes-core/packagegroups/packagegroup-tools-benchmark.bb
+++ b/meta-mel-support/recipes-core/packagegroups/packagegroup-tools-benchmark.bb
@@ -7,7 +7,7 @@ RDEPENDS_${PN} += "\
     bonnie++ \
     lmbench \
     iozone3 \
-    iperf \
+    iperf3 \
     dbench \
     tbench \
     tiobench \


drewmoseley:
@Noor-Ahsan @kergoth this seems to be necessary with the 2016.11 tools.  I ran a build using CB Lite 2016.11 and there was a build failure in iperf v2.0.5.  According to the web search, iperf v2 is considered deprecated.  The build failure is fixed in iperf v3.

For reference:
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=831180
- https://sourceforge.net/projects/iperf/

Noor-Ahsan:
Looks fine to me. @kergoth I was thinking shouldn't we move this packagegroup under meta-oe as all the benchmark recipes are located under recipes-benchmark under meta-oe

kergoth:
I wouldn't be opposed to that, as a subsequent enhancement (separate pull request). It'd make world builds easier, since it wouldn't error trying to build that packagegroup if meta-oe is excluded.

drewmoseley:
OK.  I just updated the title to drop the RFC since it sounds like we have agreement to merge this and then later move it under meta-oe.  @Noor-Ahsan over to you to merge.

PR #872, opened 2016-10-13
Fix a few archive-release hash mismatches

kergoth:


meta-mel-support/recipes-core/meta/archive-release.bb
```diff
@@ -430,7 +430,7 @@ do_prepare_release () {
 
     mv deploy/* ${DEPLOY_DIR_RELEASE}/
 }
-addtask prepare_release before do_build after do_dump_headrevs do_patch
+addtask prepare_release before do_build after do_patch
```
Noor-Ahsan: this is interesting :)

kergoth: It was a remnant of the old way we saved/restored the headrevs to handle autorev. The interesting thing is bitbake ignoring a task depending on a non-existent task rather than erroring. I think I'll open a bug in the yocto bugzilla about that. It's possible it was by design at some point, to handle cases where a task may or may not exist, but we have an addtask python api now, so conditional addtasks can be done in anonymous python and event handlers, and ignoring missing task deps could mean uncaught bugs, if the user typoed it, or forgot to inherit a class or something.

PR #882, opened 2016-11-09
psplash-*.service: adjust TMPDIR

abelal:
The psplash application checks for the TMPDIR environment
variable when starting and picks up /tmp as its temp dir
if this variable is not set. The /tmp gets mounted
pretty late in the boot process and on some faster booting
systems a failure can be seen where psplash-quit fails to
communicate with psplash-start because of a failure while
creating the communication pipe under this TMPDIR.
We now make use of the /run dir as the TMPDIR which is
available quite early in the boot process and avoids such
scenarios.
Also cleanup the psplash-quit service to drop the TimeoutSec
setting which is not applicable when the service is of type
oneshot.

### DIFF
--- a/meta-mel/recipes-core/psplash/mel/psplash-quit.service
+++ b/meta-mel/recipes-core/psplash/mel/psplash-quit.service
@@ -3,9 +3,9 @@ Description=Terminate Psplash Boot Screen
 After=psplash-start.service
 
 [Service]
+Environment=TMPDIR=/run
 Type=oneshot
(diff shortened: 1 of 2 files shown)


abelal:
@Noor-Ahsan, please review.

Noor-Ahsan:
@skrishnakar kindly review this.

skrishnakar:
@Noor-Ahsan Looks good to me. Kindly merge this. Thanks!

PR #883, opened 2016-11-09
Toaster script update to populate layers and recipes

sharidas:
This patchset has 2 commits.
First commit modifies 2 files:
a) setup-mel-builddir
b) toaster_configuration
and creates a new file toaster-layer-recipes-config.py

The change made here is to address the following:
1) Populate toaster db with the recipes and layers
2) Cosmetic change to rename the release name from
   "Local Yocto Project" to "Mentor Embedded Linux"

To populate the toaster db the data is read from:
  BUILDDIR/conf/recipes-list.txt

Second commit has changes for generating recipes-list.txt.

### DIFF
--- a/scripts/setup-mel-builddir
+++ b/scripts/setup-mel-builddir
@@ -371,5 +371,10 @@ if [ $toasterconfig -eq 1 ]; then
        cmdline_string="$cmdline_string $i"
     done
     CMD_LINE="$cmdline_string"
+    cd "${MELDIR}"
+    if [ ! -e "${BUILDDIR}/conf/recipes-list.txt" ] && [ -e "${MACHINE}/recipes-list.txt" ]; then
+        cp -f "${MACHINE}/recipes-list.txt" "${BUILDDIR}/conf/recipes-list.txt"
+    fi
+    ${MELDIR}/meta-mentor/scripts/list-recipes "$BUILDDIR/conf/recipes-list.txt"
     $layerdir/scripts/toaster_configuration "$BUILDDIR" "$MELDIR" "$BITBAKEDIR" "$CMD_LINE"
 fi
--- a/scripts/setup-mel-builddir
+++ b/scripts/setup-mel-builddir
@@ -346,5 +346,30 @@ if [ -e $BUILDDIR/conf/local.conf ]; then
 fi
 
 if [ $toasterconfig -eq 1 ]; then
-    $layerdir/scripts/toaster_configuration "$BUILDDIR" "$MELDIR" "$BITBAKEDIR"
+    machine=$(echo "$@" | awk -F" " {'print $NF'})
+    cmdline_string="$MELDIR/meta-mentor/setup-environment "
+    check_layer_option=0
+    for i in "$@"; do
+       if [ "$i" = "-t" ]; then
+          continue
+       fi
+       if [ "$i" = "-l" ]; then
+          check_layer_option=1
+          cmdline_string="$cmdline_string $i"
+          continue
+       fi
+       if [ $check_layer_option -eq 1 ]; then
+            check_layer_option=0
+            i="\"$i\""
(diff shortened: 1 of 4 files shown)


sharidas:
@Noor-Ahsan @kergoth @skrishnakar Kindly review my change

scripts/setup-mel-builddir
```diff
@@ -346,5 +346,35 @@ if [ -e $BUILDDIR/conf/local.conf ]; then
...
+            cmdline_string="$cmdline_string $i"
+            continue
+       fi
+       if [ "$i" = "$machine" ]; then
+           cmdline_string="$cmdline_string -b \"%BUILDDIR%\" $machine"
+           continue
+       fi
+       cmdline_string="$cmdline_string $i"
+    done
+    CMD_LINE=$cmdline_string
```
kergoth: You just constructed a variable with spaces, and then assigned it to another variable without quoting it. Not a good idea.

scripts/setup-mel-builddir
```diff
@@ -346,5 +346,35 @@ if [ -e $BUILDDIR/conf/local.conf ]; then
...
+          continue
+       fi
+       if [ $check_layer_option -eq 1 ]; then
+            check_layer_option=0
+            i="\"$i\""
+            cmdline_string="$cmdline_string $i"
+            continue
+       fi
+       if [ "$i" = "$machine" ]; then
+           cmdline_string="$cmdline_string -b \"%BUILDDIR%\" $machine"
```
kergoth: Here you're adding -b $BUILDDIR when you find machine, but machine is optional. If it's not found, then you'll end up not having -b passed. Is that the intent? Have you tested this when sourcing meta-mentor/setup-environment to reconfigure an existing build directory?

scripts/setup-mel-builddir
```diff
@@ -346,5 +346,35 @@ if [ -e $BUILDDIR/conf/local.conf ]; then
...
+           continue
+       fi
+       cmdline_string="$cmdline_string $i"
+    done
+    CMD_LINE=$cmdline_string
+    cd "${MELDIR}"
+    if [ ! -e "${BUILDDIR}/conf/recipes-list.txt" ] && [ -e "${MACHINE}/recipes-list.txt" ]; then
+        cp -f "${MACHINE}/recipes-list.txt" "${BUILDDIR}/conf/recipes-list.txt"
+    fi
+    ${MELDIR}/meta-mentor/scripts/list-recipes "$BUILDDIR/conf/recipes-list.txt"
```
kergoth: You're running a script (`list-recipes`) that isn't included in this pull request. Was a `git add` missed?

scripts/toaster_configuration
```diff
@@ -21,7 +21,7 @@ configured_layers () {
 }
 
 create_toaster_configuration() {
-DISTRO=$(grep "DISTRO =" "$BUILDDIR/conf/local.conf"  | awk {'print $3'} | sed -e s:\'::g)
+DISTRO=$(grep -m 1 "DISTRO =" "$BUILDDIR/conf/local.conf"  | awk {'print $3'} | sed -e s:\'::g)
```
kergoth: `grep -m` isn't mandated by posix. Please use a more portable construct. I'd recommend piping to `head -n 1`.

scripts/setup-mel-builddir
sharidas: @kergoth The build directory name will be passed by toaster. And that happens when you trigger any builds from toaster UI. So lets say if you have created a project ( consider it to be first project created in toaster ), if you trigger to build console-image, then a new build directory will be created with build-toaster-2. Similary if you create another project, then the build directory will change to build-toaster-3. Your observation is much appreciated here. So I will check again for command line build, because for UI builds we need a -b option.

kergoth: Understood, thanks for clarifying.

PR #895, opened 2016-11-17
archive-release: fix PDK_DISTRO_VERSION (quoting issue)

kergoth:


kergoth:
We were getting a literal value of `$pdk_version`, which is obviously wrong.

PR #898, opened 2016-11-22
Change demos install paths for consistency

ahsanhussain:
Move the demos from various locations to /opt/${PN} for consistency.

### DIFF
--- a/meta-mel/mel-iot/recipes-iot/libcoap/libcoap_%.bbappend
+++ b/meta-mel/mel-iot/recipes-iot/libcoap/libcoap_%.bbappend
@@ -1,0 +1,2 @@
+EXTRA_OECONF_append_mel = " --datadir=/opt"
+FILES_${PN}-examples_append = " /opt/${PN}/*"
--- a/meta-mel/recipes-graphics/mesa-demos/mesa-demos_%.bbappend
+++ b/meta-mel/recipes-graphics/mesa-demos/mesa-demos_%.bbappend
@@ -1,0 +1,2 @@
+bindir_mel = "/opt/${PN}"
+EXTRA_OECONF_append_mel = " --bindir=${bindir}"
--- a/meta-mel/qt5-mel/recipes-graphics/qtbase/qtbase_%.bbappend
+++ b/meta-mel/qt5-mel/recipes-graphics/qtbase/qtbase_%.bbappend
@@ -1,0 +1,1 @@
+OE_QMAKE_PATH_EXAMPLES_mel = "/opt/qt5/examples"
--- a/meta-mel/qt5-mel/recipes-graphics/qtdeclarative/qtdeclarative_%.bbappend
+++ b/meta-mel/qt5-mel/recipes-graphics/qtdeclarative/qtdeclarative_%.bbappend
@@ -1,0 +1,1 @@
+OE_QMAKE_PATH_EXAMPLES_mel = "/opt/qt5/examples"
--- a/meta-mel/qt5-mel/recipes-graphics/qtmultimedia/qtmultimedia_%.bbappend
+++ b/meta-mel/qt5-mel/recipes-graphics/qtmultimedia/qtmultimedia_%.bbappend
@@ -1,0 +1,1 @@
+OE_QMAKE_PATH_EXAMPLES_mel = "/opt/qt5/examples"
--- a/meta-mel/qt5-mel/recipes-graphics/qtquick1/qtquick1_%.bbappend
+++ b/meta-mel/qt5-mel/recipes-graphics/qtquick1/qtquick1_%.bbappend
@@ -1,0 +1,1 @@
+OE_QMAKE_PATH_EXAMPLES_mel = "/opt/qt5/examples"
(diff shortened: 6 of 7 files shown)


ahsanhussain:
@Noor-Ahsan, @drewmoseley, @abelal please review/merge.

@Noor-Ahsan please let me know if the changes need to be moved from meta-mentor-staging.

ahsanhussain:
@kergoth, kindly review if the current path for changes is correct.

drewmoseley:
I think we need to move the qt examples changes from https://github.com/MentorEmbedded/meta-qt5-mel/issues/109 to here.  Also, I'm not sure if meta-mel-support is the right layer or if meta-mel is better.

kergoth:
Further, adding to EXTRA_OECONF is completely pointless. We already pass all target paths to the configure script, just set the OE variables as appropriate, if necessary.

ahsanhussain:
@drewmoseley 
>I think we need to move the qt examples changes from MentorEmbedded/meta-qt5-mel#109 to here. 

Already did. It's the commit 33b66bf

ahsanhussain:
@kergoth 
> We already pass all target paths to the configure script, just set the OE variables as appropriate, if necessary.

Sorry, did not get this. What OE variables are you referring to? STAGING_{DATADIR,BINDIR} and friends?

>And why aren't we using a variable to control this change in path policy, rather than hardcoding it in each bbappend?

Because we're doing this only for some (read demos/benchmarks/tests etc) recipes? If this is not explanatory, I do not understand how to affect only a set of recipes with a variable, without any bbappends.

ahsanhussain:
@kergoth 
>     meta-mel-support is the wrong place for any bbappends. The only reason meta-mel-support exists is for new recipes to comply with yocto

Fixed now; moved the changes to meta-mel.

>    You didn't put the appends in the layer-specific area under meta-mel, so the build will break if you try to build without qt5-mel included, for example

ah yes, skipped my mind. Fixed now.

drewmoseley:
@kergoth @ahsanhussain I am certainly fine with using the FHS (presumably that means installing in /usr/bin rather than /opt/<blah>).  My main concern is consistency.  Before this work started we had a number of test apps in /usr/bin, a number in /opt/<blah> and it wouldn't surprise me if there were others.

@Noor-Ahsan @kergoth Is FHS(/usr/bin) a preferred approach here?  It will still require tech pubs and QA updates for the test apps that were previously in /opt/<blah> but that's not a fundamentally different thing.

_If_ we decide to do something non-standard then I like the idea of controlling it via a bitbake variable.

meta-mel/mel-iot/recipes-iot/libcoap/libcoap_%.bbappend
```diff
@@ -0,0 +1,2 @@
+EXTRA_OECONF_append_mel = " --datadir=/opt"
```
kergoth: This line is pointless, for example. It's also just wrong, since it could easily affect more than just examples/samples. But if you're going to take this approach, then just drop `EXTRA_OECONF` and do this:
```
datadir = "/opt"
```

As `--datadir=${datadir}` is already passed to ./configure in autotools.bbclass.

meta-mel/recipes-graphics/mesa-demos/mesa-demos_%.bbappend
```diff
@@ -0,0 +1,2 @@
+bindir_mel = "/opt/${PN}"
+EXTRA_OECONF_append_mel = " --bindir=${bindir}"
```
kergoth: This line too is pointless, as it's already passed by autotools.bbclass. Just set bindir and move on.

kergoth:
Requesting input from @drewmoseley and @ahsanhussain re: the requested changes. Thanks!

kergoth:
It's also worth noting that making sample/example paths consistent is likely something upstream would accept, if done well, in which case it's only the specific paths chosen which would be mel-specific.

drewmoseley:
libexecdir is fine by me.  The big problem with the demos that are already installed in /opt/ is that they are not in the PATH and finding them can be tricky.  If we use libexecdir, and that is by default in the PATH then that problem is resolved.

My main concern is consistency.  We should use the same mechanism for all such demos.  And if that gets us the benefit of upstream submission, then so much the better.

kergoth:
Neither /opt or libexecdir would be in the default PATH, only if they're installed in bindir would they be, but I don't think demos are particularly discoverable anyway, as you have to know what binaries to run. If the paths are made consistent, they can at least `ls /usr/lib/*/examples` or `ls /opt/*/examples`, though.

drewmoseley:
@kergoth regarding the ptest changes, can you provide a link?  I'm not clear on where/what this is.

kergoth:
https://wiki.yoctoproject.org/wiki/Ptest has the details on how it works and how it's enabled, but in a nutshell the upstream project unit tests are compiled, installed, and packaged in a separate binary package only if 'ptest' is in DISTRO_FEATURES, there's a ptest-pkgs image feature to install them all, and a ptest-runner to run the tests on target.

drewmoseley:
OK.  I agree that ptest-like functionality should be considered a longer term improvement.

@ahsanhussain please make the changes requested above and we can then decide if /opt/ or libexecdir is better.

ahsanhussain:
We decided not to do this generally, since the footprint of the change would be huge across different projects and it would be a big effort to modify each and every test/demo/benchmark/example recipe. See details on INTAMDDET-1744. Closing this PR as per decision in internal tech call discussion.

PR #900, opened 2016-11-23
update_gio_module_cache: fix postinst owner adjustment

kergoth:
We need to correct the ownership of the generated cache file, but only if a
cache file was generated (which is not the case when the modules dir is
empty).

JIRA: SB-8312, SB-8314


kergoth:
This fixes the failure to run the gio module postinst intercept for console-image builds.

PR #902, opened 2016-11-29
bbpath-intercepts.bbclass: set cleandirs on assemble_intercepts

kergoth:
Having the cleandirs on do_rootfs resulted in the path being cleaned after
assemble_intercepts was run, and an empty intercepts dir being passed to the
rootfs construction, so no intercepts were run at all. Bitbake handles
cleandirs/dirs on a per-function, not per-task, basis, so it belongs on the
prefunc.

Thanks to Abdur Rehman for investigating this and identifying the root cause.

JIRA: SB-8314

### DIFF
--- a/meta-mentor-staging/classes/bbpath-intercepts.bbclass
+++ b/meta-mentor-staging/classes/bbpath-intercepts.bbclass
@@ -20,6 +20,7 @@ python assemble_intercepts () {
     bb.utils.mkdirhier(intercepts_dir)
     bb.process.run(['cp'] + intercepts + [intercepts_dir])
 }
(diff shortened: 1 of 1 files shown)


kergoth:
If you would, please verify the fix. Thanks.

Noor-Ahsan:
@abdulrehman-MentG kindly verify this.

arehman199:
Verified this fix for mx6q and t4240. The overridden update_gio_module_cache intercept script does execute now along with the normal ones.
All postinstall intercept hooks succeed for mx6q and none for t4240(ki) as expected.

PR #903, opened 2016-11-29
gdbserver-gplv3: avoid reimplementing license whitelisting

kergoth:
oe-core has built in support for license whitelisting, so make use of it. We still
need our bits which remove gdbserver from the codebench-debug feature when
it's not buildable, however.

As a consequence of this, we now get messages from bitbake reminding us that it's being included despite the incompatible license:

    NOTE: INCLUDING gdb as buildable despite INCOMPATIBLE_LICENSE because it has been whitelisted
    NOTE: INCLUDING gdbserver-external as buildable despite INCOMPATIBLE_LICENSE because it has been whitelisted

I think this behavioral change is an improvement, since it keeps the user
aware of the exception to the rule.


kergoth:
Superceded.

PR #904, opened 2016-11-30
user_features: avoid timing-dependent variable changes

kergoth:
Rather than altering at ConfigParsed time, use _append/_remove directly.


Noor-Ahsan:
is this for dogwood. just realize PR as sent for master not morty.

kergoth:
No, it's master intentionally. This isn't a bugfix, but an improvement.

PR #905, opened 2016-12-01
python3: fix for multilib environment

aarshad:
Python module search path should not be hardcoded to lib/python

More details can be found in INTAMDDET-1724

Signed-off-by: Adeel Arshad <adeel_arshad@mentor.com>

### DIFF
--- a/meta-mel/recipes-devtools/python/python3/0001-python3-correct-the-multilib-support.patch
+++ b/meta-mel/recipes-devtools/python/python3/0001-python3-correct-the-multilib-support.patch
@@ -1,0 +1,48 @@
+From 2c3cbf6f7774a646aa322b357f30df75d1a9e89a Mon Sep 17 00:00:00 2001
+From: Adeel Arshad <adeel_arshad@mentor.com>
+Date: Wed, 16 Nov 2016 15:00:02 +0500
+Subject: [PATCH] python3: correct the multilib support
+
+When python3 rebased its multilib patch, the hard coded "lib" path
+isn't really changed because of the rebasing's error, and cause
+phthon3's failure when running on 64bit platforms as below:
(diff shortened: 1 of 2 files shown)


abelal:
@aarshad, we probably need to push this to the morty branch rather than master.

Noor-Ahsan:
if we think that this should be upstreamed then it should be in staging layer.

aarshad:
@abelal , @Noor-Ahsan , yes it is not specific to MEL and would be good to be upstreamed. so staging layer seems to be a good place holder for this. BTW what is morty branch for ?

abelal:
@aarshad, morty is the current upstream release dogwood is based on.

aarshad:
so it should go to branch morty and in directory meta-mentor-staging... any objections ?

kergoth:
I'm fairly sure this was already fixed upstream, did you check for that and look into backporting that instead?

aarshad:
@kergoth , apparently it does not seems to be fixed in python repository.
https://github.com/python/cpython/blob/3.6/Modules/getpath.c#L497

Not sure if that has been fixed by changing some other part of code, figuring that out require more investigation.

kergoth:
"[OE-core] [PATCH 3/3] python3: fix library search path" on the list fixes exactly this issue, in the same way, though is slightly more thorough. Looks like it just hasn't been merged yet.

aarshad:
Thanks @kergoth , we will try that patch and also check that if that patch also resolve the issue mentioned in
http://jira.alm.mentorg.com:8080/browse/INTAMDDET-1723
As this issue also seems relevant to the python search path configurations.

aarshad:
closing the PR as https://github.com/MentorEmbedded/meta-mentor/pull/909 provides the improved fix.

PR #906, opened 2016-12-01
mel.conf: drop MACHINE_FEATURES for PowerPC targets

Yasir-Khan:
Signed-off-by: Yasir-Khan <yasir_khan@mentor.com>

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -206,9 +206,6 @@ KERNEL_IMAGETYPES_append = " vmlinux"
 # Sourcery Analyzer with lttng-ust
 TARGET_LDFLAGS += "-Wl,--build-id=sha1"
 
-# We need vfat support for PPC targets as well
-MACHINE_FEATURES_append_powerpc = " vfat"
-
 # Explicitly remove the proprietary stuff
 MACHINE_HWCODECS = ""
 


arehman199:
@Yasir-Khan vfat is present in distro features too, also being added by mel.conf. Isn't this supposed to do it for all bsps?

Yasir-Khan:
@arehman199 In packagegroup-base recipe, packagegroup-base-vfat is added based on MACHINE_FEATURES and not on DISTRO_FEATURES.  
My understanding is that these two variables themselves do not add any packages to rootfs themselves but are used as checks in indivdual recipes to control which bits become part of build.

kergoth:
MACHINE_FEATURES controls what base packagegroups are emitted, DISTRO_FEATURES controls what base packagegroups are included automatically via packagegroup-base-extended. Only if both are set will the utils be installed in the images by default.

Yasir-Khan:
Thanks @kergoth for making it more clear. After this change "vfat" becomes part of the combined features and hence part of the image.

arehman199:
We are adding vfat to machine features using local.conf.append.\<machine\> in meta-mentor-bsp for these bsps:
- ls1
- cyclone
- minnowmax

How about we remove `MACHINE_FEATURES_append_powerpc = " vfat"` from mel.conf(we are not supporting any 32 bit ppc target atm) and append vfat to MACHINE_FEATURES in meta-mentor-bsp/t4240rdb-64b/conf/local.conf.append.t4240rdb-64b instead?

kergoth:
Removing it from mel.conf sounds good to me.

Yasir-Khan:
I've updated the PR.

PR #907, opened 2016-12-05
96boards-tools: allow udevd to process device creation events on imx6…

fahadarslan:
…ulevk-mel

resize-helper.service executes /usr/sbin/resize-helper script while udevd
is still processing events. As a result parameters like ID_PART_ENTRY_NUMBER
and ID_PART_TABLE_TYPE are not available for root device and script fails to
funciton properly. Running udevadm settle allows udevd to process device
creation events so that they are available before they are used.

Jira ID: SB-8359


Noor-Ahsan:
if it is BSP specific then why not you move it to BSP layer?

fahadarslan:
Closing this PR in favor of 
https://github.com/MentorEmbedded/meta-mx6/pull/124

PR #909, opened 2016-12-06
python3: fix two multilib issues

kergoth:
- Apply the pending oe-core patch to fix the encodings import
- Adjust site.py to search both sys.lib and 'lib', to align with the behavior
  of the python 2.7 multilib patch and fix the import of non-core modules

Closes #905
JIRA: INTAMDDET-1723


aarshad:
@mwpow3ll has confirmed that this PR resolves the issue mentioned in JIRA: INTAMDDET-1723. And it also contains the changes submitted in previous PR, so we are good to move forward and merge this.

Acked-by: Adeel Arshad <adeel_arshad@mentor.com>

PR #910, opened 2016-12-06
iperf: ensure we check for stdbool.h

kergoth:
iperf was conditionally checking for the size of bool based upon
ac_cv_header_stdbool_h, but wasn't actually calling AC_CHECK_HEADERS to make
sure that variable is set. Explicitly check for stdbool.h with
AC_CHECK_HEADERS and operate based upon that result.

Also correctly quote the first AC_DEFUN argument, while I'm here, to avoid the
warning from autoconf.

This issue was causing failures for MEL builds.

JIRA: SB-8356


kergoth:
Supercedes #58. @sharidas, please confirm on your end.

sharidas:
@kergoth Thanks a bunch for the helping hand. I have verified the same in my machine. And it works like charm.

Noor-Ahsan:
this basically supersedes https://github.com/MentorEmbedded/meta-mentor-bsp/pull/58.

PR #911, opened 2016-12-07
mel: lowercase the SANITY_TESTED_DISTROS

kergoth:
Upstream now lowercases the NATIVELSBSTRING, but not the entries in
SANITY_TESTED_DISTROS, so do so ourselves to avoid untested warnings.


kergoth:
This is for master, not morty.

PR #916, opened 2016-12-09
mel: fix taskhash mismatch for wic templates

kergoth:
After WKS_FULL_PATH is set to the generated path, but before it's been
written, the os.path.exists() will fail, resulting in a different checksum
during that period, so operate based on WKS_FULL_PATH being set instead.

JIRA: SB-8410


kergoth:
This is breaking our builds and we need them as soon as we can get them, so merging this now.

PR #922, opened 2016-12-14
Various fixes for X32

kergoth:
With this, an oe-core (no meta-oe) world build for nodistro will succeed, other than an ICE in qemu. All our supported images build fine for mel as well.

This pull request is for master, at this time. These changes can be cherry-picked into our morty branch, as needed, in the future.


Noor-Ahsan:
@kergoth  just checking the possibility that can we move this changes in https://github.com/MentorEmbedded/meta-mentor-bsp/tree/master/intel-corei7-64 instead of meta-mentor.

kergoth:
This belongs upstream, and isn't in any way specific to that machine, that's just the only one we'll be supporting. But yes, we could certainly put it there for the release. Maybe we can place it there for morty, and skip meta-mentor entirely, instead submitting the series directly to oe-core for post-morty.

kergoth:
Another option would be to put it in its own layer for the time being, as a staging area independent of meta-mentor-staging, which our bsp layer could depend upon.

PR #925, opened 2016-12-16
fix ip assignment delay issue

sbobade:
patch set to fix: ip assignment delay issue

JIRA-ID: SB-8389,SB-8351

Signed-off-by: Shrikant Bobade <shrikant_bobade@mentor.com>


sbobade:
@Noor-Ahsan  the systemd distro. feature check removed. thanks

PR #933, opened 2016-12-20
udev-extraconf: use Label ID for mounting USB mass storage

Yasir-Khan:
Use the modified udev mount.sh script file so that USB mass storage
device is mounted according to Label ID instead of device ID.

JiraID: SB-8404

Signed-off-by: Yasir-Khan <yasir_khan@mentor.com>
Signed-off-by: Sujith H <Sujith_Haridasan@mentor.com>

### DIFF
--- a/meta-mentor-staging/recipes-core/udev/udev-extraconf/mount_modified.sh
+++ b/meta-mentor-staging/recipes-core/udev/udev-extraconf/mount_modified.sh
@@ -1,0 +1,94 @@
+#!/bin/sh
+#
+# Called from udev
+#
+# Attempt to mount any added block devices and umount any removed devices
+
+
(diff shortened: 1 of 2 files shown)


kergoth:
This should have been done differently. This doesn't make it clear exactly what version of mount.sh this is based upon, nor what the delta was, so looking at this in the future it'll be decidedly non-trivial to update it to the current upstream.

The correct process for overriding a file from an upstream layer is to copy the pristine upstream file into our layer, add it and commit it as is, with a mention of exactly what commit hash was used from what upstream layer, then modify it and add and commit the modification in a second commit. Then it's both trivial to examine the diff of the changes you made to it, and is equally trivial to update the baseline and re-apply those changes to a newer upstream file.

PR #934, opened 2016-12-20
toaster-layers-recipes-config: Fix issue for Ubuntu 16.04

sharidas:
Instead of passing layer_and_recipes to iterator, it's
better to convert it into list and pass the keys explicitely.
This fixes the issue caused:

Creating toaster configuration
Traceback (most recent call last):
  File "/home/qatools/akif/mpsoc_tc3/mel/2017.02.67/meta-mentor/scripts/toaster-layers-recipes-config.py", line 43, in <module>
    for layername in layer_and_recipes:
RuntimeError: OrderedDict mutated during iteration

JIRA: SB-8480

### DIFF
--- a/scripts/toaster-layers-recipes-config.py
+++ b/scripts/toaster-layers-recipes-config.py
@@ -40,9 +40,14 @@ with open(recipeListFile, 'r') as f:
 
 # Now time to populate custom.xml
 
-for layername in layer_and_recipes:
+del_layers = []
+
+for layername in list(layer_and_recipes):
     if len(layer_and_recipes[layername]) == 1:
-        del layer_and_recipes[layername]
+        del_layers.append(layername)
+
+for layername in del_layers:
+    del layer_and_recipes[layername]
 
 pk = len(bblayers) + 1
 


sharidas:
@Noor-Ahsan @skrishnakar  @kergoth  Kindly review my change.

scripts/toaster-layers-recipes-config.py
```diff
@@ -40,7 +40,7 @@
 
 # Now time to populate custom.xml
 
-for layername in layer_and_recipes:
+for layername in list(layer_and_recipes.keys()):
```
kergoth: .keys() is redundant. Just `list(layer_and_recipes)` will work fine. This looks reasonable, though, you definitely should not be modifying something you're iterating over during the iteration :)

sharidas: @kergoth Yah, you are right. .keys() is redundant. I missed that part, that I am trying to modify dict during iteration. I will update the same in my next version of the patch. Thanks for the valuable feedback.

scripts/toaster-layers-recipes-config.py
```diff
@@ -40,9 +40,14 @@
...
-for layername in layer_and_recipes:
+del_layers = []
+
+for layername in list(layer_and_recipes):
     if len(layer_and_recipes[layername]) == 1:
-        del layer_and_recipes[layername]
+        del_layers.append(layername)
+
+for layername in del_layers:
+    del layer_and_recipes[layername]
```
kergoth: This is pointless. Converting to list() is all you needed to do in order to make it safe to delete during the iteration. Perhaps I didn't make that clear enough in the previous review. You don't modify what you're iterating. If you use list(), then you aren't modifying what you're iterating anymore, you're iterating over a list, not the dictionary.

PR #935, opened 2016-12-21
toaster_configuration: Fix for RHEL7 using virtualenv

sharidas:
There is a potential bug in the virtualenv residing in
RHEL 7.3 version. The bug is listed here:
https://github.com/pypa/virtualenv/issues/463

So inorder to overcome this issue, a work around created.

JIRA: SB-8485

### DIFF
--- a/scripts/toaster_configuration
+++ b/scripts/toaster_configuration
@@ -189,6 +189,12 @@ EOF
 }
 
 create_toaster_venv () {
+    if [ -f /etc/redhat-release ]; then
+        # If its RHEL 7 there is a bug in virtualenv which has to be tweaked
+        if [ $(cat /etc/redhat-release  | awk {'print $7'} | cut -d "." -f1) -eq 7 ]; then
+            easy_install --user --upgrade virtualenv
+        fi
+    fi
     python -m virtualenv -p python3 "$BUILDDIR/venv"
     python -m virtualenv -p python2 "$BUILDDIR/venv"
     (


sharidas:
@kergoth Kindly review my change.
Nothing else pops of my head apart from this work around :(
If you can suggest a better approach, it would help us fix this issue in a better way.

scripts/toaster_configuration
```diff
@@ -189,8 +189,26 @@ EOF
...
+                python -m virtualenv -p python3 "$BUILDDIR/venv"
+                cd "$BUILDDIR/venv/bin"
+                unlink python
+                unlink python2
+                unlink python2.7
+                cd -
+                python -m virtualenv -p python2 "$BUILDDIR/venv"
+            )
+        fi
+    else
```
kergoth: I'm assuming we need to unlink in order to let virtualenv fully update the environment after upgrading virtualenv?  Could we use --clear, instead? Alternatively, we could install the latest virtualenv into the user local python area and then use that when we construct it originally, rather than installing the new virtualenv into the virtualenv.

scripts/toaster_configuration
```diff
@@ -189,8 +189,18 @@ EOF
...
 create_toaster_venv () {
-    python -m virtualenv -p python3 "$BUILDDIR/venv"
-    python -m virtualenv -p python2 "$BUILDDIR/venv"
+    if [ $(lsb_release -i | awk {'print $3'}) == "RedHatEnterpriseWorkstation" ]; then
+        # If its RHEL 7 there is a bug in virtaulenv which has to be tweaked
+        if [ $(lsb_release -r | awk {'print $2'} | cut -d "." -f1) -eq 7 ]; then
+            easy_install --user --upgrade virtualenv
+            export PATH=$HOME/.local/bin:$PATH
+            python -m virtualenv -p python3 "$BUILDDIR/venv"
+            python -m virtualenv -p python2 "$BUILDDIR/venv"
```
kergoth: RHEL of any version other than 7 will end up not setting up the venv at all, by this logic.

scripts/toaster_configuration
```diff
@@ -189,8 +189,18 @@ EOF
...
+        fi
+    else
+        python -m virtualenv -p python3 "$BUILDDIR/venv"
+        python -m virtualenv -p python2 "$BUILDDIR/venv"
```
kergoth: The python -m lines are duplicated. Just do the easy_install & PATH adjustment in the rhel7 block and then run the python lines outside the conditional in all cases.

scripts/toaster_configuration
```diff
@@ -189,8 +189,18 @@ EOF
 }
 
 create_toaster_venv () {
-    python -m virtualenv -p python3 "$BUILDDIR/venv"
-    python -m virtualenv -p python2 "$BUILDDIR/venv"
+    if [ $(lsb_release -i | awk {'print $3'}) == "RedHatEnterpriseWorkstation" ]; then
+        # If its RHEL 7 there is a bug in virtaulenv which has to be tweaked
```
kergoth: Typo: virtaulenv

scripts/toaster_configuration
kergoth: Also, is that PATH adjustment needed at all? We aren't running 'virtualenv', we're running 'python -m virtualenv'.

scripts/toaster_configuration
```diff
@@ -189,6 +189,12 @@ EOF
 }
 
 create_toaster_venv () {
+    if [ -f /etc/redhat-release ]; then
+        # If its RHEL 7 there is a bug in virtualenv which has to be tweaked
+        if [ $(cat /etc/redhat-release  | awk {'print $7'} | cut -d "." -f1) -eq 7 ]; then
```
kergoth: Worth noting that redhat-release does exist in all redhat variants, not only rhel. I don't think it'll be an issue, since in e.g. CentOS the version number is field 3 rather than 7, but just to be aware of it.

PR #936, opened 2016-12-21
toaster-layers-recipes-config: Fix the iteration mess done

sharidas:
Removing the extra list created del_layers. Instead in the loop,
removing the elements.


sharidas:
@Noor-Ahsan @kergoth Kindly review the change.

PR #943, opened 2016-12-29
mel: also add CentOSLinux-7.* to tested distros

kergoth:


kergoth:
There are slight variations in naming depending on whether info is pulled from lsb_release or /etc/redhat_release, based on whether the lsb tools are installed.

Noor-Ahsan:
got it. Thanks.

PR #945, opened 2017-01-03
toaster: use layer names from layer.conf

abelal:
The current mechanism relies on folder names for
identification of a particular layer. This is not
suitable as folder names can be same over different
layers and tends to pollute the toaster setup as
it gets conflicting entries.
We change the behavior to make use of the real layer
name (BBFILE_COLLECTIONS) from each layer's layer.conf
which are bound to be unique for a correct bitbake
environment setup.

JIRA Ticket: SB-8582

### DIFF
--- a/scripts/toaster-layers-recipes-config.py
+++ b/scripts/toaster-layers-recipes-config.py
@@ -2,13 +2,38 @@
 
 import sys
 import os
+import re
 from collections import OrderedDict
 
 bblayers, builddir, meldir = sys.argv[1], sys.argv[2], sys.argv[3]
 
-bblayers = bblayers.split(",")
+OE_CORE_NAME = "openembedded-core"
 
-totalLayers = bblayers
+configured_layers = [OE_CORE_NAME]
+
+def get_layer_name(lpath):
+    layerconf = lpath + "/conf/layer.conf"
+    lc = open(layerconf, "r")
+    for ln in lc:
+        ln = ln.rstrip("\n")
+        layer = re.match("BBFILE_COLLECTIONS *\+= *\"", ln)
+        if layer:
+            layer = ln.split("\"")[1]
+            if layer == "core":
+                layer = OE_CORE_NAME
+            break
+    lc.close()
+    return layer
+
+# Parse through layers from bblayers.conf and pick up layer names
+for base_layer in bblayers.split():
+    layer = get_layer_name(base_layer)
+    if layer:
+        if layer == OE_CORE_NAME:
+            continue
+        configured_layers.append(layer)
+
+totalLayers = list(configured_layers)
 
 layer_and_recipes = OrderedDict()
 
@@ -20,36 +45,28 @@ layerabsPath = []
 toStrip = None
 layer = None
 
-#Populate layer_and_recipes 
-
+# Populate layer_and_recipes
 with open(recipeListFile, 'r') as f:
     for line in f:
         line = line.rstrip("\n")
+        # Check if it's a layer or a recipe
         if line.endswith("/"):
-            line = line.rstrip("/")
             toStrip = line
-            if len(line.split("/")) > 1:
-                layer = line.split("/")[-1]
-            else:
-                layer = line
-            layer_and_recipes[layer] = [line]
+            line = line.rstrip("/")
+            layer = get_layer_name(line)
+            if layer:
+                layer_and_recipes[layer] = [line]
         else:
             if line.endswith('.bb'):
-                layer_and_recipes[layer].append(line.replace(toStrip + "/", ""))
+                layer_and_recipes[layer].append(line.replace(toStrip, ""))
 
-
-# Now time to populate custom.xml
-
-for layername in list(layer_and_recipes):
-    if len(layer_and_recipes[layername]) == 1:
-        del layer_and_recipes[layername]
-
-pk = len(bblayers) + 1
-
+# Now we need to add all layers that are found in the MEL install dir
+# to custom.xml, layers from bblayers.conf are already added
+pk = len(configured_layers) + 1
 fd = open(customXML, 'a')
-
 for layername in layer_and_recipes:
-    if len(layer_and_recipes[layername]) == 0 or layername in bblayers:
+    # Skip the layer entry if it's already configured
+    if len(layer_and_recipes[layername]) == 0 or layername in configured_layers:
         continue
     totalLayers.append(layername)
     fd.write('  <object model="orm.layer" pk="%s">\n' %(str(pk)))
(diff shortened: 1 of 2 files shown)


abelal:
@Noor-Ahsan, @sharidas, please review.

scripts/toaster-layers-recipes-config.py
```diff
@@ -20,36 +39,40 @@
...
+                layer = re.match("BBFILE_COLLECTIONS *\+= *\"", ln)
+                if layer:
+                    layer = ln.split("\"")[1]
+                    if layer == "core":
+                        layer = "openembedded-core"
+                    elif not layer.startswith("meta-"):
+                        layer = "meta-" + layer
+                    layer_and_recipes[layer] = [line]
+                    break
+            lc.close()
```
kergoth: Duplicated code with line 14.

scripts/toaster_configuration
```diff
@@ -151,12 +154,15 @@ EOF
...
-   fi
-   collectLayers="$collectLayers,$layername"
-   dirpath=$(basename "$layer")
+   layername=$(grep "BBFILE_COLLECTIONS *+= *\"" ${layer}/conf/layer.conf | cut -d '"' -f2)
+   case $layername in
+       core) continue
+             ;;
+       meta-*)
+             ;;
+       *) layername="meta-${layername}"
```
kergoth: More duplicated code. Also, what's the point of artificially adding 'meta-'? Not all layers include that at all, either in the path on disk or the layer name internally.

abelal:
@kergoth, please review. I understand there's room for much improvement but I made the changes according to the current structure of the scripts which we can probably handle in the future.
Regarding the 'meta-' prefixing. It's purely aesthetic as people are used to seeing layers with such names as the layer repositories are usually named in the same manner. I can drop that if you still think it's not a valid assumption :)

sharidas:
@abelal One thing I have observed here is the layer names that get displayed in the project page( which you are already aware of ). The addition of meta before the layer name is looking bit weird. For instance meta-qt5 is the directory name and the layer name is qt5-layer. Now it looks like meta-qt5-layer.

kergoth:
Can we use the full path relative to MELDIR or TOPDIR rather than artificially creating a "layer name" from the collection name? I.e. 'meta-mentor-bsp/common' rather than either 'common' or 'meta-mentor-bsp-common'?

Noor-Ahsan:
I think we should stick to the layer name that we have defined in layer.conf. if we will take @kergoth's suggestion then we will have issue with meta-mx6 repo where we have meta-mx6-common layer insider meta-mx6. if we take only layer name from layer.conf then we can ask our customer if they wanna see which layer it is they can grep the string and they will find the layer.

abelal:
Removed "meta-" prefixing.

scripts/toaster_configuration
```diff
@@ -13,13 +13,23 @@ MELDIR="$2"
...
+
 BBLAYERS=
 configured_layers () {
    tac "$BUILDDIR/conf/bblayers.conf" | \
    sed -n -e '/^"/,/^BBLAYERS = /{ /^BBLAYERS =/d; /^"/d; p;}' | \
    awk {'print $1'}
 }
 
+get_layer_name () {
+   name=$(grep "BBFILE_COLLECTIONS *+= *\"" ${1}/conf/layer.conf | cut -d '"' -f2)
```
kergoth: This will fail, likely in a very ugly way, if there's more than one of these lines. bitbake doesn't assume only one collection per layer, that's just a convention.

abelal:
@kergoth, done.

kergoth:
Just a note that the manual extraction of BBFILE_COLLECTIONS would also fail if it was a multi-line definition. Is there a reason you didn't use bb-print-layer-data from meta-mentor, which uses python to correctly parse a layer.conf and dump its info to stdout?

abelal:
@kergoth, didn't know about the bb-print-layer-data, pretty cool :)
Anyhow, if I try and use bb-print-layer-data both in the shell script (toaster_configuration) and the other python script (toaster-layers-recipes-config.py) the time for setting up the environment increases quite a lot.
with bb-print-layer-data

> real	0m45.799s
> user	0m32.827s
> sys	0m2.521s

with manual parsing

> real	0m21.021s
> user	0m5.311s
> sys	0m1.108s

What do you suggest we do here? I believe timing isn't that much of a problem because it will be a one time setup but using bb-print-layer-data everywhere would make maintenance easy as we'd be using the same approach everywhere.
I applied the following patch on top of the changes here to make use of bb-print-layer-data, do you see any obvious mistakes which would explain the timing difference?
```
diff --git a/scripts/toaster-layers-recipes-config.py b/scripts/toaster-layers-recipes-config.py
index 350ec8b..958b3a8 100755
--- a/scripts/toaster-layers-recipes-config.py
+++ b/scripts/toaster-layers-recipes-config.py
@@ -3,6 +3,7 @@
 import sys
 import os
 import re
+import subprocess
 from collections import OrderedDict
 
 bblayers, builddir, meldir = sys.argv[1], sys.argv[2], sys.argv[3]
@@ -12,17 +13,9 @@ OE_CORE_NAME = "openembedded-core"
 configured_layers = [OE_CORE_NAME]
 
 def get_layer_name(lpath):
-    layerconf = lpath + "/conf/layer.conf"
-    lc = open(layerconf, "r")
-    for ln in lc:
-        ln = ln.rstrip("\n")
-        layer = re.match("BBFILE_COLLECTIONS *\+= *\"", ln)
-        if layer:
-            layer = ln.split("\"")[1]
-            if layer == "core":
-                layer = OE_CORE_NAME
-            break
-    lc.close()
+    layer = subprocess.check_output(meldir + "/meta-mentor/scripts/bb-print-layer-data " + lpath + " 2> /dev/null | cut -d ':' -f1", shell=True).decode('ascii').strip()
+    if layer == "core":
+        layer = OE_CORE_NAME
     return layer
 
 # Parse through layers from bblayers.conf and pick up layer names
diff --git a/scripts/toaster_configuration b/scripts/toaster_configuration
index c6d82eb..7e98a72 100755
--- a/scripts/toaster_configuration
+++ b/scripts/toaster_configuration
@@ -23,7 +23,7 @@ configured_layers () {
 }
 
 get_layer_name () {
-   name=$(grep "BBFILE_COLLECTIONS *+= *\"" ${1}/conf/layer.conf | head -n 1 | cut -d '"' -f2)
+   name=`${MELDIR}/meta-mentor/scripts/bb-print-layer-data ${1} 2> /dev/null | cut -d ':' -f1`;
    if [ "$name" = "core" ]; then
       name="$OE_CORE_NAME"
    fi
```

kergoth:
The performance is definitely a concern, though as you say a relatively minor one. IIRC bb-print-layer-data can accept multiple arguments, and outputs the data for multiple layers, perhaps you could gather the info in the caller of get_layer_name, instead, so we can batch executions of the script and thereby avoid the delay of the python vm startup?

I'd be fine with sticking with manual parsing for now, merging this, and then improving it later, to prevent this being a further bottleneck.

abelal:
@Noor-Ahsan, @kergoth, thanks.

PR #949, opened 2017-01-05
mel: only add wic.bmap, not wic.bz2

kergoth:
Plenty of machines are now adding wic to IMAGE_FSTYPES themselves. We don't
want or need to duplicate that, and IMAGE_FSTYPES is generally in the
machine's control anyway, so let them do it. Keep wic.bmap, as that's more of
a distro choice, and doesn't result in duplication.

JIRA: SB-8529

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -54,9 +54,9 @@ LOGO_pn-plymouth = "${PLYMOUTH_LOGO}"
 IMAGE_FSTYPES ?= "tar.bz2 ext3"
 UBI_VOLNAME = "rootfs"
 
-# If a wks file exists for this BSP, use it to emit a wic image
+# If a wks file exists for this BSP, also write a .bmap file for bmaptool
 WKS_FULL_PATH ??= ""
-IMAGE_FSTYPES += "${@'wic.bz2 wic.bmap' if '${WKS_FULL_PATH}' else ''}"
+IMAGE_FSTYPES += "${@'wic.bmap' if '${WKS_FULL_PATH}' else ''}"
 
 # Additional dependencies for deployed files are often pulled in via
 # do_image_wic[depends], to ensure the files are available for


Noor-Ahsan:
I'll merge it on Monday after testing on each BSP and updating boot isntructions.

Noor-Ahsan:
@kergoth we were thinking over here that why not we add wic as well in mel.conf as we are adding wic.bmap as wic.bmap need wic file.

kergoth:
Because mel is the distro. The distro does *not* decide how images are constructed, the machine does that via IMAGE_FSTYPES. Generating the .bmap is about the distro, as it's up to the distro whether it wants to support bmaptool.

kergoth:
Also, read the description of the pull request: Plenty of machines are now adding wic to IMAGE_FSTYPES themselves. We don't want or need to duplicate that, and IMAGE_FSTYPES is generally in the machine's control anyway, so let them do it.

Noor-Ahsan:
Jsut for my understanding what value we will have if we haven't set wic in IMAGE_FSTYPES and wic.bmap is added?

kergoth:
I don't understand the question. If you look at the line:
```
IMAGE_FSTYPES += "${@'wic.bmap' if '${WKS_FULL_PATH}' else ''}"
```

wic.bmap is only added when we're going to be writing a wic image. If the machine sets WKS_FILE to a valid path and doesn't add wic to IMAGE_FSTYPES, that's a bug in the machine.

Noor-Ahsan:
got it. Thanks.

kergoth:
Better yet would be to only add wic.bmap when wic or wic.<compression format> is in IMAGE_FSTYPES, rather than using WKS_FULL_PATH, but we can't add anonymous python to IMAGE_FSTYPES which references IMAGE_FSTYPES, as that'd be recursion. The only way to add it to IMAGE_FSTYPES based on IMAGE_FSTYPES would be an event handler or anonymous python, which seemed like excessive complexity, since it'd have to be split out into a .inc or .bbclass.

PR #950, opened 2017-01-06
Warn and disable rootfs resizing for non-GPLv3 builds

kergoth:
Rootfs resizing relies on a script which needs recent parted, which is GPLv3,
so we can't build in non-GPLv3 without disabling it or reworking the script.

JIRA: SB-8334

### DIFF
--- a/meta-mel/classes/resize-rootfs-gplv3.bbclass
+++ b/meta-mel/classes/resize-rootfs-gplv3.bbclass
@@ -1,0 +1,25 @@
+# Warn the user and disable rootfs resizing for non-GPLv3 configurations
+#
+# We use two events, as it seems ConfigParsed is firing multiple times in
+# a build at the moment.
+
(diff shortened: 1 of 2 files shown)


kergoth:
Whether we care about supporting non-GPLv3 builds enough to risk impacting our QA cycle I leave to someone else to decide :)

kergoth:
Note that we can provide a workaround / document a known issue without merging this, by instructing the user to remove the addition of 96boards-tools from conf/local.conf.

Noor-Ahsan:
@chombourger what is your opinion on this?

chombourger:
I have no issues with this fix going in - many customers require non GPLv3 so I am glad we caught this. In fact our first Dogwood customer is DensoWave and we have to make sure they can build non GPLv3 images. Fortunately we can get them to use fixed size images

PR #955, opened 2017-01-13
bmap-tools: Downgrade bmap-tools recipe to version 2.5

sharidas:
We are downgrading bmap-tools recipe from the version
available at poky, which is 3.2. The reason is it generates
wic.bmap file with version 2 which is not supported in Ubuntu
14.04. So if we keep version 2.5 then it works on Ubuntu 14.04
and 16.04.

JIRA: SB-8621

### DIFF
--- a/meta-mel-support/recipes-support/bmap-tools/bmap-tools_2.5.bb
+++ b/meta-mel-support/recipes-support/bmap-tools/bmap-tools_2.5.bb
@@ -1,0 +1,24 @@
+SUMMARY = "Tools to generate block map (AKA bmap) and flash images using bmap"
+DESCRIPTION = "Bmap-tools - tools to generate block map (AKA bmap) and flash images using \
+bmap. Bmaptool is a generic tool for creating the block map (bmap) for a file, \
+and copying files using the block map. The idea is that large file containing \
+unused blocks, like raw system image files, can be copied or flashed a lot \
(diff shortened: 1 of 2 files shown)


sharidas:

Noor-Ahsan:
@kergoth  kindly review

kergoth:
Have we looked at the differences here? What functionality is lost in the reversion? What exactly caused the file format change?

Noor-Ahsan:
@kergoth no we did not have time for that. we just checked it working.

skrishnakar:
@Noor-Ahsan @kergoth  The below listed are the key differences that came in v3.2 since 2.5 release:

1. 10% faster reading and decompressing using more parallelism.
2. Multi-stream bzip2 archives are now supported.
3. LZO archives are now supported.
4. Make 'bmaptool create' work with the "tmpfs" file-system.
5. Decompression should now require less memory, which should fix out-of-memory problems reported  by some users on machines with 2GB RAM.
6. Support multi-stream bzip2 archives, like the ones created by the "pbzip2" tool.
7. Switch from using SHA1 checksums in the bmap file to SHA256.
8. Support OpenPGP (AKA gpg) signatures for the bmap file.
9. The Fiemap module now always synchronizes the image before scanning it for mapped areas.
10. On-the-fly decompression support for '.xz' and '.tar.xz' files.

PR #967, opened 2017-01-24
dosfstools: Apply the patch for specific version of dosfstools

sharidas:
dosfstools has both GPLv3 recipe as well as GPLv2 recipe in poky.
If user sets INCOMPATIBLE_LICENSE = "GPLv3" the build was failing
because GPLv2 version of the recipe was picked and the patch failed
to apply for the same. This patch renames the file from dosftools_%.bbappend
to dosfstools_4.0.bbappend.

JIRA: SB-8683


sharidas:

skrishnakar:
This looks reasonable! Based on verification of http://jira.alm.mentorg.com:8080/browse/SB-8456 on both versions 2.11 & 4.0. Thanks @sharidas !

PR #968, opened 2017-01-24
Sb 8671

arehman199:


arehman199:
@Noor-Ahsan This updates the Freescale-EULA licensed recipes(eth-config and fm-ucode) in meta-mentor/meta-mel/fsl-ppc.

I noticed that rcw recipe in meta-fsl-ppc was also freescale-eula licensed in dogwood so added a compiled binary for it as well. This was BSD licensed in cedar.
The rcw recipe in meta-fsl-arm was(in cedar) and still is(in dogwood) BSD licensed so no need to add a compiled binary for ls1.

PR #971, opened 2017-01-27
Use an event handler to add wic.bmap to IMAGE_FSTYPES

kergoth:
If a wic image type is enabled, also enable wic.bmap. We use an event handler,
as we can't reference IMAGE_FSTYPES from inline python in the same variable,
and while we can reference WKS_FULL_PATH, then the value in the config
metadata would be different than in image recipes. This wouldn't normally be
a problem, but toaster is using the config metadata value.

JIRA: SB-8685

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -54,9 +54,8 @@ LOGO_pn-plymouth = "${PLYMOUTH_LOGO}"
 IMAGE_FSTYPES ?= "tar.bz2 ext3"
 UBI_VOLNAME = "rootfs"
 
-# If a wks file exists for this BSP, also write a .bmap file for bmaptool
-WKS_FULL_PATH ??= ""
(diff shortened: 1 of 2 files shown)


kergoth:
@Noor-Ahsan, I leave it to you whether to do anything with this pull req or not, but it's here if we need it. I verified both wic and wic.bmap are in IMAGE_FSTYPES in the config metadata with this change applied.

kergoth:
I still think ideally toaster should be pulling the variable from the metadata for the image in question, as that's what's actually used to build the image, but this would work for now.

Noor-Ahsan:
@kergoth! @sharidas  is in discussion with toaster community on this. Probably it will be resolved in coming days.

Noor-Ahsan:
I am going to merge your PR untill we get solution from toaster.

kergoth:
Fine with me, thanks @Noor-Ahsan

PR #984, opened 2017-02-17
Custom xml consolidation

abelal:
- Unifies the logic for creating custom.xml.
- Minor cleanups and clarifications.

JIRA Ticket: SB-8614.

### DIFF
--- a/scripts/toaster-layers-recipes-config.py
+++ b/scripts/toaster-layers-recipes-config.py
@@ -9,8 +9,6 @@ builddir, meldir, cmdline = sys.argv[1], sys.argv[2], sys.argv[3]
 
 OE_CORE_NAME = "openembedded-core"
 
-configured_layers = [OE_CORE_NAME]
-
 def get_layer_name(lpath):
     layerconf = lpath + "/conf/layer.conf"
     lc = open(layerconf, "r")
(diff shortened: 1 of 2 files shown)


abelal:
@Noor-Ahsan, please review.

scripts/toaster-layers-recipes-config.py
```diff
@@ -25,27 +23,129 @@ def get_layer_name(lpath):
...
+        machine_found = True
+    if distro_found and machine_found:
+        break
+lc.close()
 
-layer_and_recipes = OrderedDict()
-
-recipeListFile = builddir + "/conf/recipes-list.txt"
 customXML = builddir + "/custom.xml"
+fd = open(customXML, 'a')
```
sharidas: Wouldn't it be nice if we make it as write mode instead of append mode?

sharidas: The change looks pretty neat to me. And I appreciate that you took the effort to consolidate most of the work into the python file. Thanks for the work @abelal  :)

abelal: @sharidas, very valid and would make it more clear where the file is actually created at first. I'll make the change.

PR #986, opened 2017-02-23
setup-environment/-mel-builddir: exclude mentor-private on mel-lite

abelal:
meta-mentor-private only includes stuff related to ADE and
PDK license verification both of which are not supported
on MEL Lite so various issues are seen.
We now disable inclusion of meta-mentor-private in bblayers
rather than adjusting each feature according to DISTRO so
it can be maintained much cleanly.


abelal:
@kergoth, @Noor-Ahsan, please review.

@Noor-Ahsan, I hope this implements what you had in mind :)

Noor-Ahsan:
@abelal yes

PR #993, opened 2017-03-29
mel.conf: make mentor-staging inherits conditional

kergoth:
This fixes the ability to include meta-mel without meta-mentor-staging.


kergoth:
I ran into this when attempting to perform yocto compliance testing on our layers with both the new upstream layer check tool and my own.

PR #994, opened 2017-03-31
Revert "setup-{rh,ubuntu}: drop 32-bit libs"

abelal:
CB Lite is a 32 bit toolchain and we provide installers for
MEL Lite with the same so these are still needed.

This reverts commit 2e3857c5c75ee5606bfe5352d9863863a9e3a2f9.

JIRA Ticket: SB-8954.

### DIFF
--- a/scripts/setup-rh
+++ b/scripts/setup-rh
@@ -8,7 +8,7 @@
 
 packages="make gcc gcc-c++ patch texi2html diffstat texinfo tetex cvs git
           subversion gawk tar gzip bzip2 redhat-lsb sqlite ncurses-devel \
-          SDL-devel glibc-devel glibc-static \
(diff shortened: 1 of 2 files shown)


abelal:
@kergoth, @skrishnakar, @wfarnsworth, please review.

abelal:
Thinking a bit more on this. We can probably document this in the MEL Lite quick start guide rather than doing it this way but that seems like a marketing decision to me.

wfarnsworth:
@abelal I'd prefer to handle it this way, but I will defer to the decision of @kergoth and @skrishnakar 

Acked-by: Wade Farnsworth \<wade_farnsworth@mentor.com\>

wfarnsworth:
@Noor-Ahsan: please review

skrishnakar:
@abelal  This looks fine to me. Thanks!

PR #997, opened 2017-04-11
Add various missing flex/bison deps

kergoth:


kergoth:
Spotted these while testing a world build with pkgconf.

kergoth:
elm, no point holding off on the merge, there's no real risk, and the fixes are no brainers.

PR #999, opened 2017-05-04
wic-bmap.inc: append wic.bmap to IMAGE_FSTYPES instead of setting _ap…

frederiksdun:
…pend

    and initramfs typically sets IMAGE_FSTYPES to INITRAMFS_FSTYPES and does not use
    WIC. If _append is used wic.bmap is appended and will fail on empty WKS_FULL_PATH


frederiksdun:
The original change was for toaster.
Could anyone test this change for toaster?

Noor-Ahsan:
@frederiksdun BTW for which release you need this change.

frederiksdun:
I'll add this our staging area for dogwood, but it should be fixed for next  (major) release.

PR #1001, opened 2017-05-22
setup-mel-builddir: fix tty handling for the prompt

kergoth:
Once we're in prompt_choice, we can no longer test stdin or stdout to see if
it's a tty, since stdin is used to feed the prompt choices, so test for it
outside the function and pass an argument to force non-interactive mode when
appropriate.


kergoth:
This issue was breaking the prompt when multiple BSPs were installed.

PR #1002, opened 2017-05-22
Implement git repository delivery for MEL

kergoth:
This should not be merged until #1001 and #1000 are merged, as it depends upon those.

### DIFF
--- a/meta-mel-support/recipes-core/meta/archive-release.bb
+++ b/meta-mel-support/recipes-core/meta/archive-release.bb
@@ -29,7 +29,7 @@ ARCHIVE_RELEASE_DL_TOPDIR ?= "${DL_DIR}"
 def configured_update_layers(d):
     """Return the configured layers whose basenames are update-*"""
     update_layers = set()
-    for layer in d.getVar('BBLAYERS', True).split():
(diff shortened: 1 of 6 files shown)


kergoth:
Updated to split back up mel-checkout and the sourced setup script, to avoid shell compatibility issues. Moved mel-checkout into a scripts dir at mel install toplevel, and renamed the toplevel setup-environment back to the name it had when they were merged, setup-mel, which is a more appropriate name given what it does now. Also reworked how we tell mel-checkout to disable extra manifest inclusion to be more clean.

kergoth:
Hold off a bit on merging this, as I'm prepping a couple more manual builds of this for @cp-mentor before we get our automated builds working with it.

Noor-Ahsan:
@kergoth what is the status of this.

kergoth:
I had to split these changes off into a separate layer to test with dogwood/morty, and had to fix a number of issues I ran into with it, so now I'm pulling those changes back into meta-mentor and cleaning up and rebasing and will update the branch at that time, then we can get it merged.

kergoth:
This should be good to go, now.

PR #1008, opened 2017-06-08
Initramfs

fahadarslan:


fahadarslan:
Related to 
https://stash.alm.mentorg.com/projects/MEL/repos/mel-meta-ultrazed-eg-iocc/pull-requests/4/overview

Noor-Ahsan:
@fahadarslan double check the branch where this should be be merged. i think this should be on dogwood-asycn1 branch

Noor-Ahsan:
Its fine. it should be on morty.

kergoth:
'customize for hypervisor team' is worse than useless as a commit message. Exactly what is needed, why, and why is it always appropriate for our builds by default? There's no way for anyone looking at this commit, either its log message or its change to the bbappend to be able to tell exactly what was changed in the package list, without examining the original recipe for comparison.

PR #1009, opened 2017-06-13
release/setup-environment: fix passing of -b to setup-env

kergoth:
- If we pass `-b .` after `$@`, then it won't be obeyed when a MACHINE is
  explicitly specified, as getopts won't handle options after arguments
- Since we moved `-b .` to before `$@`, we need to ensure that any explicitly
  passed `-b` isn't passed along to the underlying setup-environment script


kergoth:
@Noor-Ahsan, this fixes the issues you ran into with the example output earlier.

PR #1016, opened 2017-07-05
core-image-minimal-initramfs: Provide openssh to intiramfs.

Noor-Ahsan:
* Provided openssh package by adding ssh-server-openssh in IMAGE_FEATURS


kergoth:
This says *what* it does, but not *why*.

PR #1018, opened 2017-07-27
chkconfig-alternatives*: fully obey the sysroot for removeLinks

kergoth:
JIRA: SB-9676


kergoth:
The facility was adjusted to be relative to the sysroot when creating the links, but not when removing them.

mwpow3ll:
Looks good; builds for me

Acked-by: Michael Powell <michael_powell@mentor.com>

PR #1022, opened 2017-09-06
packagegroup-tools-benchmark: Remove dbench and tbench in case of non…

Noor-Ahsan:
… GPLV3.

* We only have GPLV3 version of tbench and dbench. So when we set
  INCOMPATIBLE_LICENSE = "GPLv3" this recipe complaint that tbench and
  dbench has GPLV3 license so this can not be built. Removed these packages
  from packagegroup in case of incompatible license set to GPLv3.


kergoth:
Looks good, thanks.

PR #1030, opened 2017-10-17
setup-mel-builddir: use ${TOPDIR} in BBLAYERS

kergoth:
This improves readability, by using ${TOPDIR} rather than a full absolute path
for layers relative to TOPDIR. It should ideally also adjust paths relative to
COREBASE/OEROOT/MELDIR, but that's less of a concern given the new mel
installer setup, and adding more generic determination of relative paths would
add more complexity to the scripts.

### DIFF
--- a/scripts/setup-mel-builddir
+++ b/scripts/setup-mel-builddir
@@ -150,7 +150,7 @@ configure_for_machine () {
         echo >&2 "Error in execution of bb-determine-layers, aborting"
         exit 1
     fi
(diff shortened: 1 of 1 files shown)


kergoth:
I.e. will show `${TOPDIR}/poky/meta` rather than `/path/to/mel.../poky/meta`

Noor-Ahsan:
@kergoth  do we need this after we moved to oe-core+bitbake

kergoth:
This has nothing to do with poky vs oe-core/bitbake, it has to do with whether we include full absolute paths in BBLAYERS or the paths relative to the root of TOPDIR (which is the same as the root of the mel layers in the new delivery).

PR #1031, opened 2017-10-26
scancode.bbclass: [Request for review]: OI-2 Adding bbclass for scancode

hmsadiq:
This is adding the functionality to run scancode utility on code.
You need to set the following variables to use it. You can override
 these from local.conf or auto.conf.
1) SCANCODE_GIT_LOCATION: Git Path for scancode-toolkit cloning, default is set.
2) SCANCODE_TAG: Select tag, if you want.
3) SCANCODE_FORMAT: Output file format json or html-app.
4) SCANCODE_SRC_LOCATION: Location to store scancode-toolkit.

This belongs to OI-2.

Signed-off-by: H M Irfan Sadiq <irfan_sadiq@mentor.com>

### DIFF
--- a/meta-mentor-staging/classes/scancode.bbclass
+++ b/meta-mentor-staging/classes/scancode.bbclass
@@ -1,0 +1,35 @@
+# We use scancode utlity for extacting licence information.
+# scancode itself is an OSS Utlitity.
+# For more informaiton https://github.com/nexB/scancode-toolkit
+
+SCANCODE_FORMAT ?= "html-app"
+EXT = "${@'html' if d.getVar('SCANCODE_FORMAT', True) == 'html-app' else 'json'}"
+SCANCODE_TOOLKIT = "${@get_scancode_toolkit(d)}"
+SCANCODE_TAG = "v2.2.1"
+SCANCODE_GIT_LOCATION ?= "https://github.com/nexB/scancode-toolkit.git"
+SCANCODE_SRC_LOCATION ?= "${DL_DIR}/scancode"
+
+def get_scancode_toolkit(d):
+    lf = bb.utils.lockfile(d.getVar('SCANCODE_SRC_LOCATION', True) + ".lock")
+    if (not os.path.exists(d.getVar('SCANCODE_SRC_LOCATION', True))):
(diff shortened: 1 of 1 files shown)


meta-mentor-staging/classes/scancode.bbclass
```diff
@@ -0,0 +1,35 @@
...
+	cd ${SCANCODE_TOOLKIT}
+	if [ -d "${S}" ]; then
+		./scancode ${S} --format  ${SCANCODE_FORMAT} ${DEPLOY_DIR_IMAGE}/scancode/${PN}.${EXT}
+	fi
+}
+
+addtask scancode after do_patch
+
+do_scancode_oss() {
+    echo "We are done running scancode"
```
Noor-Ahsan: Isn't BBNOTE is better than echo and also print the package name

hmsadiq:
- I have tried using recipe to for source fetching which presents following challenges

1. Adding its DEPENDS in bbclass causes circular dependency which tried to limit it so that it is added only for the target recipes as follows.

>    
    deps = ""
    if not d.getVar('INHIBIT_DEFAULT_DEPS', False):
        if (d.getVar('HOST_SYS', True) != d.getVar('BUILD_SYS', True)):
            deps += " scancode-native "
    return deps
2. scancode is not very simple package, it itself has a lot of dependencies if we need to install it on native sysroot, initially it is failing for some python-native-mail module.
3. Even if we are able to satisfy these dependencies, it has some thirdparty packages as well for successful launch.

Also, similar tools are not installed natively like fossology/spdx http://www.pelagicore.com/using-yocto-and-fossology-to-get-spdx-licence-output/

Do you have any suggestions how should we proceed with it ?
May be, we have these options
- Add class as it is.
- remove download part altogether from the class and leave installation of the tool on user and assume they will set location variable.
- Add class as it is to some QA layer and keep it internal.

kergoth:
I'd suggest adding it to HOSTTOOLS and rely on the user having it installed on the host. Keeping the download portion is not going to work, so either we need a functioning native recipe, or rely on it as a host tool.

hmsadiq:
HOSTTOOLS variable is not available in dogwood hence replacing this PR with https://github.com/MentorEmbedded/meta-mentor/pull/1057

PR #1032, opened 2017-11-08
mel: remove the dependence upon poky

kergoth:
- Drop the dep from meta-mel on meta-poky/meta-yocto
- Copy the bits from poky.conf rather than including it

JIRA: SB-10324


kergoth:
See also https://github.com/MentorEmbedded/mel-manifest/pull/523.

PR #1035, opened 2017-11-28
setup-environment: add layer search paths argument

mdurnev:
From https://github.com/MentorEmbedded/adit-custom/pull/580

### DIFF
--- a/scripts/setup-mel-builddir
+++ b/scripts/setup-mel-builddir
@@ -9,7 +9,7 @@
 MEL_DISTRO="${MEL_DISTRO:-mel}"
 CORELAYERS="${CORELAYERS:-core mel-support mentor-staging}"
 if [ "${MEL_DISTRO}" != "mel-lite" ]; then
-    OPTIONALLAYERS="${OPTIONALLAYERS:-mentor-private}"
+    OPTIONALLAYERS="${OPTIONALLAYERS-mentor-private}"
 fi
 EXTRAMELLAYERS="${EXTRAMELLAYERS:-openembedded-layer filesystems-layer networking-layer multimedia-layer sourcery gplv2}"
 layer_priority_overrides="openembedded-layer=1"
--- a/scripts/setup-mel-builddir
+++ b/scripts/setup-mel-builddir
@@ -41,6 +41,10 @@ Options:
   -b BUILDDIR  Specify the build directory to create (default: 'build')
   -o MELDIR    Specify the root of the MEL install, where you can find bitbake
                and the layers
+  -p PATHS     Specify colon separated list of paths, where you can find layers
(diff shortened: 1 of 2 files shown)


scripts/setup-mel-builddir
```diff
@@ -41,6 +41,8 @@ Options:
   -b BUILDDIR  Specify the build directory to create (default: 'build')
   -o MELDIR    Specify the root of the MEL install, where you can find bitbake
                and the layers
+  -a PATHS     Specify colon separated list of additional directories, where
+               you can find layers
```
kergoth: This should indicate that wildcards are valid, in some fashion, and might also want to mention what directories are searched/checked by default -- i.e. "what is this in addition *to*?"

scripts/setup-mel-builddir
```diff
@@ -124,7 +126,7 @@ configure_for_machine () {
     echo 'BBLAYERS = "\' >> $BUILDDIR/conf/bblayers.conf
     layersfile=$(mktemp setup-mel-builddir.XXXXXX)
 
-    layerpaths="*:*/*:meta*/*/*:$MELDIR/*:$MELDIR/*/*:$MELDIR/meta*/*/*"
+    layerpaths="*:*/*:meta*/*/*:$MELDIR/*:$MELDIR/*/*:$MELDIR/meta*/*/*${DIRS:+:}$DIRS"
```
kergoth: `DIRS` is 1) too vague a name, 2) inconsistent with the case of our existing commandline argument variables, and 3) isn't set to a default value along with our other commandline option variables, meaning its value would also be picked up from the user's environment if set, which is particularly risky given the generic variable name. Please name it something more appropriate and set a default value, thanks!

mdurnev:
@kergoth, thank you very much for your comments. The PR has been updated. Is it better now?

Noor-Ahsan:
@kergoth are you still planning to use -a option for this. have you considered using different option as we discussed on IRC

kergoth:
Having thought about it, I agree that -a isn't the most intuitive choice here, I think -L as you suggested (-L for search path, -l for what to search for, like the linker) or -p (for path, rather than -a for additional) are better options. I'll adjust it before I merge, so @mdurnev doesn't need to mess with this further. Thanks for the reminder.

mdurnev:
@kergoth, I'd like https://github.com/MentorEmbedded/adit-custom/pull/580 to be in sync with this PR. I want to use the same option in adit-custom. Just let me know which option to use, and I can modify both PRs.

kergoth:
Understood. Let's go with -p. As it supports and uses wildcards, the semantics differ from that of -L, which could be confusing.

mdurnev:
@kergoth, the PR has been updated.

PR #1041, opened 2017-12-22
local.conf.append: Introduce MEL_ENABLE_JTAG_SUPPORT flag

fahadusman:
Adding this flag to common local.conf.append in meta-mel for future use, in Dogwood-Async3, this flag would be added via bsp-specific local.conf.append files.


kergoth:
I don't think this is a very good idea. If the bsp specific local.conf.append is going to add this anyway, then what's the point in adding something that won't be supported here? *If* this is something that requires core setup amongst the bsps, then we need a sanity check to warn the user if they enable it here but the bsp doesn't support it. I'd suggest looking at how we're handling swupdate for one example

PR #1044, opened 2018-01-01
dhcp: fix build issue with libxml2 support

abelal:
A missing case breaks the build when libxml2 is
required and found appropriately. The third argument
to the function AC_SEARCH_LIB is action-if-found which
was mistakenly been used for the case where the library
is not found and hence breaks the configure phase
where it shoud actually pass.
We now pass on silently when action-if-found is
executed.
Already accepted upstream: http://cgit.openembedded.org/openembedded-core/commit/?id=a17f3ec910366e9e7551fa24fbc07929b9584341

JIRA: SB-10251


abelal:
@Noor-Ahsan, please review.

PR #1048, opened 2018-01-16
setup-mel improvements

kergoth:
- Support --help
- Set CODEBENCH_PATH by default again (regression)
- Refactor to reduce the amount of work done in the sourced script, to make it easier to avoid cluttering up the user's environment

JIRA: SB-10033, SB-10034


kergoth:
I recommend reviewing on a per-commit basis, rather than as one large diff, due to the rename and refactoring.

kergoth:
@chombourger, @Noor-Ahsan, the -h/--help info for setup-mel currently only shows arguments explicitly handled by setup-mel, not those passed along to the underlying setup-environment script in meta-mentor, unless the build dir already exists, i.e. -l to add layers. This is because we can have multiple mel versions installed in the same location, and those could technically have modified setup-environment in meta-mentor, and before we've unpacked a particular version, we don't have access to the underlying setup-environment script. The only way to add those arguments would be to hardcode the ones we know exist for all versions into the usage info for setup-mel, or emit details about the supported arguments at archive-release-time for setup-mel to use to determine which arguments are common to all releases, but that seems excessive. Thoughts?

chombourger:
should we add the -l option to the --help output? I am asking because it is a commonly used option and I am somehow failing to remember that it expects a white separated list of layers (instead of e.g. multiple -l options or a single -l but with a comma separated list). I understand this option may be handled in a sub-script though

chombourger:
otherwise LGTM :)

kergoth:
Yeah, I was thinking the same, I think -l is unlikely to go anywhere, and potentially -b as well to specify the build directory, and ignore the rest. I'll go that route, thanks.

Noor-Ahsan:
@kergoth the layer search path option that Mikhail added. Will that come in this script or we need to add that somewhere else

kergoth:
@Noor-Ahsan , the intent here was to only include the most commonly used arguments, and which are likely to be supported in every mel release for a given install destination, not all of them, since we're having to hardcode them in the setup-mel script. Do we think that argument is likely to be very common amongst our customers?

Noor-Ahsan:
@kergoth I think that is a valuable command line option which we want to keep

kergoth:
Fair enough, I'll add it. Thanks for the feedback.

PR #1051, opened 2018-01-29
setup-{environment,mel-builddir}: use :- not - until we rework the logic

kergoth:
These vars are always set by setup-environment right now, so changing to
- resulted in not including layers we want to include by default. I intend to
shortly refactor setup-environment similar to how I recently refactored
setup-mel, we can fix this issue after that.


kergoth:
This fixes the issue where meta-sourcery wasn't being included by default.

Noor-Ahsan:
thanks @kergoth

PR #1056, opened 2018-02-15
Fixed overlapping mounts causing unmount and installation failures

Arsalan-Awan:
The automounter udev script attempts to mount the block devices under
/run/media/partition-label. If two devices have same partition-label,
they will be mounted on the same location.

This causes two types of known issues:

1. MEL installation fails if a MEL image is already installed on a
media other than where attempted to install such as HDD/eMMC/SSD/USB
that is attached during installation.

2. Unmount fails on a device if attempted using the device node
e.g. "/dev/sda1" with the 'umount' command.

These changes fix such issues.

Associated issues: INTAMDDET-2258, INTAMDDET-2259

Signed-off-by: Arsalan-Awan <Arsalan_Awan@mentor.com>

### DIFF
--- a/meta-mentor-staging/recipes-core/udev/udev-extraconf/mount_modified.sh
+++ b/meta-mentor-staging/recipes-core/udev/udev-extraconf/mount_modified.sh
@@ -19,9 +19,16 @@ done
 
 automount() {	
 	name="`basename "$DEVNAME"`"
-        if [ -z ${ID_FS_LABEL} ]; then
-		#Lets use UUID as the LABEL, as it will be unique for the $DEVNAME
-		ID_FS_LABEL=`/sbin/blkid | grep $DEVNAME | awk {'print $2'} | tr '=|"' ' ' | awk {'print $2'}`
+
+	# Get the LABEL or PARTLABEL
+        LABEL=`/sbin/blkid | grep "$DEVNAME" | grep -o 'LABEL=".*"' | cut -d '"' -f2`
+        # If the $DEVNAME does not have a LABEL or a PARTLABEL
+        if [ -z "$LABEL" ]; then
+                # Set the mount location dir name to $name e.g. sda (that would keep it unique)
+                ID_FS_LABEL="${name}"
+        else
+                # Set the mount location dir name to LABEL appended with $name e.g. sda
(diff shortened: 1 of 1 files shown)


Arsalan-Awan:
@Noor-Ahsan, @abelal, please review

meta-mentor-staging/recipes-core/udev/udev-extraconf/mount_modified.sh
```diff
@@ -19,10 +19,17 @@ done
 
 automount() {	
 	name="`basename "$DEVNAME"`"
-        if [ -z ${ID_FS_LABEL} ]; then
-		#Lets use UUID as the LABEL, as it will be unique for the $DEVNAME
-		ID_FS_LABEL=`/sbin/blkid | grep $DEVNAME | awk {'print $2'} | tr '=|"' ' ' | awk {'print $2'}`
```
abelal: @Arsalan-Awan, as discussed earlier I believe this technique is not accurate as well so please identify that in the commit message.

meta-mentor-staging/recipes-core/udev/udev-extraconf/mount_modified.sh
```diff
@@ -19,10 +19,17 @@ done
...
-        fi
+
+	# If the $DEVNAME has a LABEL or a PARTLABEL
+	if /sbin/blkid | grep "$DEVNAME" | grep "LABEL"; then
+		# Get the LABEL or PARTLABEL
+		LABEL=`/sbin/blkid | grep "$DEVNAME" | sed -e 's/.*LABEL=\"//' -e 's/\".*//'`
```
abelal: How about we do this before the if condition and use the LABEL variable to evaluate the if condition?

abelal: For evaluation of LABEL prior to the if condition you can probably use something like
`/sbin/blkid | grep "$DEVNAME" | grep -o 'LABEL=".*"' | cut -d '"' -f2`

Arsalan-Awan:
@abelal, please review now. I have updated my commit. Thanks

PR #1057, opened 2018-02-20
scancode.bbclass: [Request for review]: OI-2 Adding bbclass for scancode

hmsadiq:
This class is adding the functionality to run scancode utility on code.
You can override SCANCODE_FORMAT if you need json output. It assumes scancode
is present in PATH and is set in HOSTTOOLS.

This belongs to OI-2.

Signed-off-by: H M Irfan Sadiq <irfan_sadiq@mentor.com>

### DIFF
--- a/meta-mentor-staging/classes/scancode.bbclass
+++ b/meta-mentor-staging/classes/scancode.bbclass
@@ -1,0 +1,24 @@
+# We use scancode utlity for extacting licence information.
+# scancode itself is an OSS Utlitity.
+# For more informaiton https://github.com/nexB/scancode-toolkit
+# scancode is used from HOSTTOOLS.
+
+SCANCODE_FORMAT ?= "html-app"
+EXT = "${@'html' if d.getVar('SCANCODE_FORMAT', True) == 'html-app' else 'json'}"
+
+do_scancode() {
+	mkdir -p ${DEPLOY_DIR_IMAGE}/scancode
+	if [ -d "${S}" ]; then
+		scancode ${S} --format  ${SCANCODE_FORMAT} ${DEPLOY_DIR_IMAGE}/scancode/${PN}.${EXT}
+	fi
+}
+
+addtask scancode after do_patch
+
+do_scancode_oss() {
+    echo "We are done running scancode"
+}
+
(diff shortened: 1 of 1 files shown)


meta-mentor-staging/classes/scancode.bbclass
```diff
@@ -0,0 +1,24 @@
...
+
+addtask scancode after do_patch
+
+do_scancode_oss() {
+    echo "We are done running scancode"
+}
+
+do_scancode_oss[recrdeptask] = "do_scancode_oss do_scancode"
+do_scancode_oss[nostamp] = "1"
+addtask do_scancode_oss after do_scancode
```
kergoth: The do_scancode_oss task can be removed. Bitbake supports running a specific task against all dependencies recursively on its own now: `bitbake --runall=fetch console-image`.

meta-mentor-staging/classes/scancode.bbclass
```diff
@@ -0,0 +1,24 @@
...
+# For more informaiton https://github.com/nexB/scancode-toolkit
+# scancode is used from HOSTTOOLS.
+
+SCANCODE_FORMAT ?= "html-app"
+EXT = "${@'html' if d.getVar('SCANCODE_FORMAT', True) == 'html-app' else 'json'}"
+
+do_scancode() {
+	mkdir -p ${DEPLOY_DIR_IMAGE}/scancode
+	if [ -d "${S}" ]; then
+		scancode ${S} --format  ${SCANCODE_FORMAT} ${DEPLOY_DIR_IMAGE}/scancode/${PN}.${EXT}
```
kergoth: The convention is to no longer write directly to DEPLOY_DIR_IMAGE, but instead allow the sstate state management code to keep track of what it's installed and uninstall to/from that location. I'd suggest doing the same here, following the same pattern as that used by deploy.bbclass.

kergoth: Upon further consideration, I'd also suggest moving this to its own subdirectory of DEPLOY_DIR, rather than cluttering up DEPLOY_DIR_IMAGE with content that isn't directly bound to an image.

PR #1058, opened 2018-02-22
setup-environment: further improvements

kergoth:
These are general improvements, but which improve functionality in non-MEL/non-OE setups like isar.


kergoth:
This restores the -/:- fix we had to revert due to unforeseen issues, fixing it properly.

PR #1068, opened 2018-03-29
Add initial production-image, development-image

kergoth:
Also adds and uses image-sanity-incompatible.bbclass, which may be
upstream-suitable, depending on interest.

Further improvements are likely, for example a default non-root user, prevent
root login at console, etc.


Noor-Ahsan:
looks good to me

PR #1069, opened 2018-03-30
ConsoleKit: ensure fs is still mounted when logging stop/restart

bboccoqq:
/var/volatile could have disappeared, and that is where ConsoleKit logs
its start/stop/restart entries.


kergoth:
90% certain this belongs upstream and isn't mel-specific. Consider moving this to meta-mentor-staging when we work on elm.

PR #1077, opened 2018-04-19
initrdscripts/init-live.sh: Fixed mounts fail to move to real root fs

ArsalanHAwan:
When there are spaces in the mount points of devices e.g.:

 a partition mounted at "/run/media/My Root Partition-sda1",

the initrd fails to move such mount points over to the
corresponding directories at /media under the real root filesystem,
and the mount points would appear at the same location as they were
mounted on when detected by initrd, for example:
 here: "/run/media/My Root Partition-sda1"
 instead of here: "/media/My Root Partition-sda1"

This causes issues such as:

  * The disks/partitions cannot be formated with any filesystem
    using e.g. mkfs.ext4 or mke2fs in general. When tried to do so
    by making sure the device is not mounted, it failed with
    errors such as:

    > /dev/sda1 is apparently in use by the system; will not make a
      filesystem here!
    > /dev/sda1: Device or resource busy while setting up superblock

  * The read/write operations become extremely slow. e.g. Under testing,
    it took approx. 2 hours just to copy 700 MB of data to the partition,
    and it took more than 40 minutes to delete that data from it.
    Same operations took under 5 minutes on a partition that had no
    spaces in its mount point (or which was successfully moved to real
    root by initrd and appeared under /media instead of /run/media).

This commit fixes such issues by quoting the arguments of failing mount
move commands and by parsing OCT or HEX encoded special characters
such as spaces to ASCII characters in the mount points.

Associated issue: INTAMDDET-2361

### DIFF
--- a/meta-mentor-staging/recipes-core/initrdscripts/files/0001-initrdscripts-init-live.sh-Fixed-mounts-fail-to-move.patch
+++ b/meta-mentor-staging/recipes-core/initrdscripts/files/0001-initrdscripts-init-live.sh-Fixed-mounts-fail-to-move.patch
@@ -1,0 +1,65 @@
+From f8d1c573897825f8ffcc68263b9379ca9c2f775c Mon Sep 17 00:00:00 2001
+From: "Arsalan H. Awan" <Arsalan_Awan@mentor.com>
+Date: Fri, 20 Apr 2018 19:06:48 +0500
+Subject: [PATCH] initrdscripts/init-live.sh: Fixed mounts fail to move to real
+ root fs
+
+When there are spaces in the mount points of devices e.g.:
+
+ a partition mounted at "/run/media/My Root Partition-sda1",
+
+the initrd fails to move such mount points over to the
+corresponding directories at /media under the real root filesystem,
+and the mount points would appear at the same location as they were
+mounted on when detected by initrd, for example:
+ here: "/run/media/My Root Partition-sda1"
+ instead of here: "/media/My Root Partition-sda1"
+
+This causes issues such as:
+
+  * The disks/partitions cannot be formated with any filesystem
+    using e.g. mkfs.ext4 or mke2fs in general. When tried to do so
+    by making sure the device is not mounted, it failed with
+    errors such as:
+
+    > /dev/sda1 is apparently in use by the system; will not make a
+      filesystem here!
+    > /dev/sda1: Device or resource busy while setting up superblock
+
+  * The read/write operations become extremely slow. e.g. Under testing,
+    it took approx. 2 hours just to copy 700 MB of data to the partition,
+    and it took more than 40 minutes to delete that data from it.
+    Same operations took under 5 minutes on a partition that had no
+    spaces in its mount point (or that was successfully moved to real
+    root by initrd and appeared under /media instead of /run/media).
+
+This commit fixes such issues by quoting the arguments of failing mount
+move commands and by parsing OCT or HEX encoded special characters
+such as spaces to ASCII characters in the mount points.
+
+Signed-off-by: Arsalan H. Awan <Arsalan_Awan@mentor.com>
+---
+ init-live.sh | 7 +++++--
+ 1 file changed, 5 insertions(+), 2 deletions(-)
+
+diff --git a/init-live.sh b/init-live.sh
+index 441b41c9d6..e58df12061 100644
+--- a/init-live.sh
++++ b/init-live.sh
+@@ -91,8 +91,11 @@ boot_live_root() {
+     # Move the mount points of some filesystems over to
+     # the corresponding directories under the real root filesystem.
+     for dir in `awk '/\/dev.* \/run\/media/{print $2}' /proc/mounts`; do
+-        mkdir -p  ${ROOT_MOUNT}/media/${dir##*/}
+-        mount -n --move $dir ${ROOT_MOUNT}/media/${dir##*/}
++        # Parse any OCT or HEX encoded chars such as spaces
++        # in the mount points to actual ASCII chars
++        dir=`printf $dir`
++        mkdir -p "${ROOT_MOUNT}/media/${dir##*/}"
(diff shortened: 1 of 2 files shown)


ArsalanHAwan:
@Noor-Ahsan , @abelal : please review.

Noor-Ahsan:
@fahadarslan kindly review

fahadarslan:
@Noor-Ahsan, seems fine to me.

kergoth:
This doesn't seem right to me, personally. Spaces in filenames and directories are completely valid and allowed by POSIX. If there's a problem with them in the scripts processing them, those scripts should be fixed to quote their arguments.

ArsalanHAwan:
@kergoth : You are absolutely right! Quoting the arguments would be the right thing to do to fix such a problem. I have made the changes. Please review now.

meta-mentor-staging/recipes-core/initrdscripts/files/0001-initrdscripts-init-live.sh-Fixed-mounts-fail-to-move.patch
```diff
@@ -0,0 +1,65 @@
...
++++ b/init-live.sh
+@@ -91,8 +91,11 @@ boot_live_root() {
+     # Move the mount points of some filesystems over to
+     # the corresponding directories under the real root filesystem.
+     for dir in `awk '/\/dev.* \/run\/media/{print $2}' /proc/mounts`; do
+-        mkdir -p  ${ROOT_MOUNT}/media/${dir##*/}
+-        mount -n --move $dir ${ROOT_MOUNT}/media/${dir##*/}
++        # Parse any OCT or HEX encoded chars such as spaces
++        # in the mount points to actual ASCII chars
++        dir=`printf $dir`
```
kergoth: `$dir` isn't quoted here, nor is the output of the printf. Also, printf can accept multiple arguments, and format strings. What happens if '%s' is in the volume name, as unlikely as that seems?

meta-mentor-staging/recipes-core/initrdscripts/files/0001-initrdscripts-init-live.sh-Fixed-mounts-fail-to-move.patch
```diff
@@ -0,0 +1,65 @@
...
+ 1 file changed, 5 insertions(+), 2 deletions(-)
+
+diff --git a/init-live.sh b/init-live.sh
+index 441b41c9d6..e58df12061 100644
+--- a/init-live.sh
++++ b/init-live.sh
+@@ -91,8 +91,11 @@ boot_live_root() {
+     # Move the mount points of some filesystems over to
+     # the corresponding directories under the real root filesystem.
+     for dir in `awk '/\/dev.* \/run\/media/{print $2}' /proc/mounts`; do
```
kergoth: for dir in ``some command`` is going to split the output of `some command` into words and iterate over that, unless the script alters IFS. If awk is returning multiple lines, it'd be better to either alter IFS to iterate line by line, or better, use a different paradigm for the iteration:
```
awk '/\/dev.* \/run\/media/{print $2}' /proc/mounts | while read -r dir; do
    : # something with dir
done
```

ArsalanHAwan: 1. The mount point locations in /proc/mounts do not contain ASCII spaces. The spaces are rather octal escaped as '\040'. e.g.: "/run/media/My\040Root\040Partition"
2. awk's {print $2} will make sure that only one word is extracted with no included spaces at all... it would only extract e.g. similar to "/run/media/My\040Root\040Partition" in each line.
3. `for dir in` will be iterating on mount points as mention above... which is exactly what we want!

meta-mentor-staging/recipes-core/initrdscripts/files/0001-initrdscripts-init-live.sh-Fixed-mounts-fail-to-move.patch
ArsalanHAwan: 4. printf expects its first argument to be the FORMAT string, and then, **separated by spaces**, the ARG1 ARG2 ARG3 ... In our case, the `awk {print $2}` in for loop will make sure that no space separated args ever reach the printf! Therefore, quotes around `$dir` become redundant.
5. The output of printf is enclosed in the backtick/grave-accents `, as allowed by POSIX, and is directly assigned to variable `dir`. Is it necessary to quote it? In my experiance, I never had to...
6. Fancy format string entities such as '%s' in volume labels, if they ever show up, will be replaced with NOTHING since ARG1 ARG2 etc will be empty-strings/NOTHING. We don't-want/should-not-support such entities in the mount points anyway because they are only going to cause trouble everywhere!

meta-mentor-staging/recipes-core/initrdscripts/files/0001-initrdscripts-init-live.sh-Fixed-mounts-fail-to-move.patch
kergoth: Is it truly guaranteed that there will *never* be a mount in /run/mounts/ on a path with spaces? That sounds like a dangerous assumption to make assuming it'll never get fixed elsewhere in the future, particularly if we don't document that this code makes that assumption.

ArsalanHAwan: BTW, the same rule is followed in the /etc/fstab as well... we cannot write mount points in fstab with ASCII spaces. Spaces have to be octal escaped for such mounts to work!

kergoth: fstab isn't the only method by which mounts occur, but fair enough. Thanks for the thorough replies. Looks okay to me.

PR #1080, opened 2018-04-20
consolekit: add dependency on var-volatile.mount for restart/stop

bboccoqq:
consolekit restart and stop loggers try to write to /var/volatile/log
before the target completes shutdown. Add an After= dependency on
var-volatile.mount unit so that they run before during shutdown.

To achieve this, these services must be started at boot time rather than
just before shutdown, so that the order of start and stop is respected.
Instead of running the required functionality on ExecStart it must then
be run on ExecStop. In addition, since there is no dependency on
reboot/halt, a script finds out what shutdown path is run and invokes
the correct cosolekit utility. Only one service is thus needed, which
handles all cases.


bboccoqq:
Could not reproduce, but this dependency was clearly missing.

meta-mel/recipes-support/consolekit/consolekit/add-dependency-on-var-volatile.mount-for-shutdown.patch
```diff
@@ -0,0 +1,184 @@
...
+-		$(LN_S) ../console-kit-log-system-stop.service )
+-	( cd $(DESTDIR)$(systemdsystemunitdir)/reboot.target.wants && \
+-		rm -f console-kit-log-system-restart.service && \
+-		$(LN_S) ../console-kit-log-system-restart.service )
+-	( cd $(DESTDIR)$(systemdsystemunitdir)/kexec.target.wants && \
+-		rm -f console-kit-log-system-restart.service && \
+-		$(LN_S) ../console-kit-log-system-restart.service )
++	( cd $(DESTDIR)$(systemdsystemunitdir)/basic.target.wants && \
++		rm -f console-kit-log-system-shutdown.service && \
++		$(LN_S) ../console-kit-log-system-shutdown.service )
```
abelal: @bboccoqq, wouldn't it be more appropriate to use the Install section in the service so it can be enabled/disabled using systemctl on the target? This is more like
[Install]
WantedBy="basic.target"

meta-mel/recipes-support/consolekit/consolekit/add-dependency-on-var-volatile.mount-for-shutdown.patch
```diff
@@ -0,0 +1,184 @@
...
+-ExecStart=@sbindir@/ck-log-system-stop
+diff --git a/data/console-kit-shutdown-handler b/data/console-kit-shutdown-handler
+new file mode 100644
+index 0000000..8f89802
+--- /dev/null
++++ b/data/console-kit-shutdown-handler
+@@ -0,0 +1,26 @@
++#!/bin/bash
++
++last_shutdown_info_file=/var/log/ConsoleKit/last_shutdown_info
```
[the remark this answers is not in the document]
bboccoqq: Debugging purposes.

meta-mel/recipes-support/consolekit/consolekit_%.bbappend
```diff
@@ -1,5 +1,10 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 SRC_URI_append = " \
-    file://0001-Ensure-filesystems-are-still-mounted-when-consolekit.patch \
+    file://add-dependency-on-var-volatile.mount-for-shutdown.patch \
 "
+
+do_install_append() {
+    install -d ${D}${sbindir}
+    install -m 755 ${S}/data/console-kit-shutdown-handler ${D}${sbindir}
```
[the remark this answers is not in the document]
bboccoqq: Not intended for upstream, since this issue is specific to having /var/log disappearing before being able to write to it on shutdown.
So, you mean it would be better to just let the makefile do what it was doing, then removing the files it creates through the recipe instead of patching the makefile ?

meta-mel/recipes-support/consolekit/consolekit/add-dependency-on-var-volatile.mount-for-shutdown.patch
bboccoqq: This is the way the package is written, I did not try to change that behaviour. I don't think there is an expectation that the service is to be enabled/disabled at runtime.

bboccoqq:
Added commit to handle unmounting /var/volatile before the bind mounts to different directories present on /var/volatile.

meta-mel/recipes-support/consolekit/consolekit/add-dependency-on-var-volatile.mount-for-shutdown.patch
abelal: Considering that the expectation is not to have something that is runtime configurable through systemctl I believe this is fine.

meta-mel/recipes-support/consolekit/consolekit/add-dependency-on-var-volatile.mount-for-shutdown.patch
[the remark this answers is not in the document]
abelal: Understood!

meta-mel/recipes-support/consolekit/consolekit_%.bbappend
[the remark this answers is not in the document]
abelal: @bboccoqq, yup I believe that would be a more feasible approach and would allow anyone looking at the changes easily understand as the changes will be happening in a single place i.e. the recipe.

abelal:
Oh, https://github.com/MentorEmbedded/meta-mentor/pull/1080/commits/5dab00fb21094ed39f6666960917b9c11bf012b5 looks like the real dependency that would handle the race condition that occurs every now and then.

meta-mel/recipes-support/consolekit/consolekit_%.bbappend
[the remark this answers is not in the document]
bboccoqq: OK, re-implemented this way. You're right it's cleaner.

I decided to patch the package/Makefile the first time around since it was taking care of consuming the *.service.in files.

meta-mel/recipes-support/consolekit/consolekit_%.bbappend
```diff
@@ -1,5 +1,36 @@
...
+
+    # install "shutdown" service and auto-start softlink
+    sed -e 's,@sbindir@,${sbindir},g' ${WORKDIR}/${shutdown_service}.in > ${WORKDIR}/${shutdown_service}
+    install -d ${D}${systemd_system_unitdir}
+    install -m 644 ${WORKDIR}/${shutdown_service} ${D}${systemd_system_unitdir}
+
+    install -d ${D}${systemd_system_unitdir}/basic.target.wants
+    pushd ${D}${systemd_system_unitdir}/basic.target.wants
+    ln -s ../${shutdown_service}
+    popd
```
abelal: Rather than using pushd and popd I believe this can be handled in a one liner where the ln command would change to
`ln -sf ../${shutdown_service} ${D}${systemd_system_unitdir}/basic.target.wants/`

meta-mel/recipes-support/consolekit/consolekit/console-kit-shutdown-handler
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
abelal: @kergoth, is it a good idea to use bash explicitly or should we lean towards /bin/sh?

kergoth: Generally #!/bin/sh is preferred unless bashisms are truly required. If the recipe already depends/rdepends on bash it'd be less of a concern, but we don't want to pull bash into a build if we don't actually need it.

abelal: @bboccoqq, what do you think we should do here? Does your script have any bashisms?

bboccoqq: I always write scripts using bash, so I wasn't sure if I had bashisms in there, but I ran it through a checker it complained about the use of -o in two tests at the end.

Reworked those lines, invokes #!/bin/sh, remove RDEPENDS_${PN] = "bash".

PR #1088, opened 2018-05-17
volatile-binds: fix service disable in read-only file system

abdulkarim17:
systemd manages services by maintaining an elaborate tree under /etc/systemd.
When /etc/systemd is mounted as read-only, user cannot disable or enable services,
this caused error when opkg-configure tried to disable itself after first boot.
This fix mounts /etc/systemd in read-only mode allowing services to be disabled
or enabled at runtime.

Jira: http://jira.alm.mentorg.com:8080/browse/SB-11200

### DIFF
--- a/meta-mel/conf/local.conf.append
+++ b/meta-mel/conf/local.conf.append
@@ -14,7 +14,5 @@ IMAGE_FEATURES_DISABLED_PRODUCTION ?= "${IMAGE_FEATURES_DEVELOPMENT} ssh-server-
 # build gets its own identifier, so is self-contained already.
 # Default: ${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}
 #ADE_IDENTIFIER ?= "${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}.customized"
-SYSTEMD_BINDS = "\
-/var/volatile/systemd /etc/systemd\n\
- "
-VOLATILE_BINDS_append = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'read-only-rootfs', d.getVar('SYSTEMD_BINDS', True), '', d)}"
+IMAGE_CLASSES += "remove-systemd-bind"
+ROOTFS_POSTPROCESS_COMMAND_append = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'read-only-rootfs', '','remove_systemd_bind;', d)}"
--- a/meta-mel/conf/local.conf.append
+++ b/meta-mel/conf/local.conf.append
@@ -14,3 +14,7 @@ IMAGE_FEATURES_DISABLED_PRODUCTION ?= "${IMAGE_FEATURES_DEVELOPMENT} ssh-server-
 # build gets its own identifier, so is self-contained already.
 # Default: ${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}
 #ADE_IDENTIFIER ?= "${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}.customized"
+SYSTEMD_BINDS = "\
+/var/volatile/systemd /etc/systemd\n\
+ "
+VOLATILE_BINDS_append = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'read-only-rootfs', d.getVar('SYSTEMD_BINDS', True), '', d)}"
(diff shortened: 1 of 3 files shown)


meta-mentor-staging/recipes-core/volatile-binds/volatile-binds.bbappend
```diff
@@ -0,0 +1,5 @@
+# mount /etc/systemd with rw rights.
+BIND_DIR += "\
+    /var/volatile/systemd /etc/systemd\n\
+"
+VOLATILE_BINDS += "${@bb.utils.contains('IMAGE_FEATURES', 'read-only-rootfs', '${BIND_DIR}', '', d)}"
```
kergoth: IMAGE_FEATURES is image level by definition, it can be set in the image recipe itself, not just globally, so you can't access that here. Also, volatile-binds is only installed when doing read only builds, so this is just redundant. Drop the conditional entirely, or better yet, replace it with a conditional on the system distro feature.

abdulkarim17: I have dropped this changes. Kindly review again.

meta-mel/conf/local.conf.append
```diff
@@ -14,3 +14,7 @@ IMAGE_FEATURES_DISABLED_PRODUCTION ?= "${IMAGE_FEATURES_DEVELOPMENT} ssh-server-
 # build gets its own identifier, so is self-contained already.
 # Default: ${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}
 #ADE_IDENTIFIER ?= "${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}.customized"
+SYSTEMD_BINDS = "\
+/var/volatile/systemd /etc/systemd\n\
+ "
+VOLATILE_BINDS_append = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'read-only-rootfs', d.getVar('SYSTEMD_BINDS', True), '', d)}"
```
kergoth: This is still both not appropriate and not necessary. As I said in the previous review, volatile-binds is only installed on read only images, so this conditional is pointless, and it needs to be conditional on the systemd distro feature.

kergoth: As I said on IRC: it's not appropriate to muck with image features globally. It's completely valid to have one image be read-only-rootfs and another image non-read-only-rootfs in a single bitbake build, which would not work properly the way you're doing things. Absolute worst case we can use a rootfs postprocessing hook to alter the configuration after rootfs construction, based on the image feature, the way a number of other read only rootfs tweaks are made already.

If you grep oe-core for read-only-rootfs, you'll see the only time anything uses it is in image context. Recipe contents never change based on image features.

The difficulty with trying to do anything like this via a rootfs hook is that volatile-binds mounts are set up via individual systemd services, based on a template in volatile-binds SRC_URI, which is something  we won't easily get to in an image. Given this, I'd suggest arranging (likely via bbappend) to unconditionally add the needed entry to VOLATILE_BINDS, but disable the service generated for it by default, then enable it in a rootfs postprocessing hook only when read-only-rootfs and the systemd distro feature are both enabled.

meta-mel/conf/local.conf.append
```diff
@@ -14,7 +14,5 @@ IMAGE_FEATURES_DISABLED_PRODUCTION ?= "${IMAGE_FEATURES_DEVELOPMENT} ssh-server-
 # build gets its own identifier, so is self-contained already.
 # Default: ${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}
 #ADE_IDENTIFIER ?= "${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}.customized"
-SYSTEMD_BINDS = "\
-/var/volatile/systemd /etc/systemd\n\
- "
-VOLATILE_BINDS_append = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'read-only-rootfs', d.getVar('SYSTEMD_BINDS', True), '', d)}"
+IMAGE_CLASSES += "remove-systemd-bind"
+ROOTFS_POSTPROCESS_COMMAND_append = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'read-only-rootfs', '','remove_systemd_bind;', d)}"
```
kergoth: The fact that this is in meta-mel and the volatile-binds change is in meta-mentor-staging is problematic. We don't want one of these changes without the other.

kergoth:
Looks fine otherwise, though.

kergoth:
IMO it's not upstream-appropriate as is, making /etc/systemd writable when it's using non-volatile tmpfs doesn't make much sense, the changes would just revert on next reboot, which defeats the purpose of systemctl enable or systemctl disable. it only really makes sense where it's being mounted on a user data partition with our swupdate setup, IMO, so the changes persist.

Given this, I think it only really makes sense putting it in meta-mel or an appropriate swupdate layer, and potentially also making it conditional on our user data feature.

kergoth:
Merged as is for now, please submit a follow-up pull request to make the requested changes. Thanks.

PR #1089, opened 2018-05-17
X11: Remove x11 configuration from mel distro

sajjad238:
Signed-off-by: Sajjad Ahmed <sajjad_ahmed@mentor.com>

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -279,6 +279,7 @@ XSERVER ??= "\
 INITRD_IMAGE_LIVE ?= "core-image-minimal-install-initramfs"
 
 # Additional package groups
+FEATURE_PACKAGES_graphics        ?= "weston weston-init weston-examples"
 FEATURE_PACKAGES_codebench-debug ?= "gdbserver strace openssh-sftp-server"
 FEATURE_PACKAGES_nas             ?= "packagegroup-role-nas"
 FEATURE_PACKAGES_print-server    ?= "packagegroup-role-print-server"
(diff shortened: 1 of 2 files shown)


meta-mel/conf/distro/include/mel.conf
```diff
@@ -7,6 +7,9 @@ POKY_DEFAULT_DISTRO_FEATURES = "largefile opengl ptest multiarch wayland vulkan"
 POKY_DEFAULT_EXTRA_RDEPENDS = "packagegroup-core-boot"
 POKY_DEFAULT_EXTRA_RRECOMMENDS = "kernel-module-af-packet"
 
+# X11 is no longer used as display manager [replaced by wayland]
+DISTRO_FEATURES_DEFAULT_remove =  "x11"
```
kergoth: It may be better to copy and modify DISTRO_FEATURES_DEFAULT rather than using _remove, as it's not possible to undo an _remove. If we do it like this, it'd be impossible for a user to add x11 back in via local.conf, or even via their own distro that includes mel.conf as a starting point.

sajjad238: @kergoth you mean we can put DISTRO_FEATURES_DEFAULT as constant here? like
DISTRO_FEATURES_DEFAULT="acl alsa argp bluetooth ext2 ..." 
That way we've to keep track of variable updates if changed

kergoth: You're right that it wouldn't be ideal, but better that than to do something the user can't undo. I have a better suggestion: use the _REMOVE  indirection we've used elsewhere in mel.conf:
```
DISTRO_FEATURES_REMOVE ?= "x11"
DISTRO_FEATURES_remove = "${DISTRO_FEATURES_REMOVE}"
```
Then the _REMOVE variable could be altered to change what's removed.

PR #1093, opened 2018-05-25
postinst-intercepts.bbclass: include search candidates in file checksums

kergoth:
This ensures that adding an intercept to a higher priority layer will change the checksums.


kergoth:
This is the first step toward getting this upstream, as including candidate paths in a case like this is critical, which is why `bb.utils.which()` does so as well.

PR #1095, opened 2018-05-28
volatile-binds: fix service disable in read-only file system

abdulkarim17:
systemd manages services by maintaining an elaborate tree under /etc/systemd.
When /etc/systemd is mounted as read-only, user cannot disable or enable services,
this caused error when opkg-configure tried to disable itself after first boot.
This fix mounts /etc/systemd in read-only mode allowing services to be disabled
or enabled at runtime.

Jira: http://jira.alm.mentorg.com:8080/browse/SB-11200


Noor-Ahsan:
this is same PR as https://github.com/MentorEmbedded/meta-mentor/pull/1088 but for dogwood release.

PR #1097, opened 2018-06-08
meta-mel: added Multimedia as image feature.

arshadaleem66:


kergoth:
Is gstreamer all that's needed? What about the plugin groups? Also, what about hwcodecs, which often includes gstreamer plugins provided by the MACHINE? Is that pulled in implicitly elsewhere?

Noor-Ahsan:
@kergoth thanks for your input. we will look into this in separate issue http://jira.alm.mentorg.com:8080/browse/SB-11453

PR #1098, opened 2018-06-08
sdk_auto_relocate: run both aspects of enable/disable at the same time

kergoth:
There was a mismatch between when the logic which disables if lib is
empty (packaging time) and when the variables that disable it are obeyed
(parse time). Ensure that they run at the same time to resolve this
mismatch and ensure the reloc is fully disabled appropriately in all
cases where it should be.

JIRA: SB-10686


Noor-Ahsan:
@kergoth the should  be master or sumo?

kergoth:
sumo, well spotted, thanks. Apparently I need to adjust my submission scripts/aliases :)

PR #1099, opened 2018-06-11
mel: fall back from wayland to x11 based on vendor support

kergoth:
JIRA: SB-11414

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -3,7 +3,11 @@ DISTRO_VERSION[vardepsexclude] = "DATE"
 SDK_VERSION[vardepsexclude] = "DATE"
 
 # Override these in poky based distros
-POKY_DEFAULT_DISTRO_FEATURES = "largefile opengl ptest multiarch wayland vulkan"
+POKY_DEFAULT_DISTRO_FEATURES = "largefile opengl ptest multiarch vulkan"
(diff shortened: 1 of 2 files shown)


kergoth:
We indicate what the vendor supports with `MACHINE_FEATURES`, specifically the `x11` and `wayland` features. If neither is present, we use wayland as usual.

kergoth:
The last piece is to actually add x11 to `MACHINE_FEATURES` in meta-mentor-bsp for any vendors which support x11 but not wayland.

abelal:
Great! this is going to help with the AMD BSPs for sure. They aren't planning to move to wayland just yet.

kergoth:
After more testing we'll want to settle on the exact list of x11 packages to include. Right now it includes x11-base, x11-sato, and hwcodecs (though iirc that's empty for some machines due to mel.conf mucking with the hwcodecs variable).

abelal:
@kergoth, @Noor-Ahsan, we do need this for the sumo branch as well.

Noor-Ahsan:
@abelal  thanks for catching it. Yes we need it on sumo.

kergoth:
Will there be an issue opened to track the fix against sumo?

Noor-Ahsan:
Well we can test it under http://jira.alm.mentorg.com:8080/browse/SB-11229 and send a PR on sumo branch

PR #1100, opened 2018-06-13
recipetool: add 'edit' subcommand

kergoth:
This edits the recipe and any bbappends for the specified target.


kergoth:
Trivial command, but very useful, and upstream (at least Ross) has indicated interest in this going to oe-core, hence the inclusion in meta-mentor-staging.

PR #1102, opened 2018-06-13
Handle incompatible license whitelisting in perf & lvm2

kergoth:


kergoth:
@Noor-Ahsan, in case you're curious, this is how you do this, using a function in a class we have, to account for whitelisting. I.e. this lets you do a gplv3-incompatible build that still allows man to build, so perf still builds fine:
```
INCOMPATIBLE_LICENSE = "GPL-3.0 LGPL-3.0"
WHITELIST_GPL-3.0 += "man libpipeline"
```

I'll be submitting the functions from the class to oe-core's oe.license module. The bbappends to handle non-gplv3 will likely need to go upstream to meta-gplv2, however.

kergoth:
Updated to only check libpipeline to add the man rdep, as man-db itself is gplv2, its dependency, libpipeline, is what's gplv3. Though I may just drop the man rdep entirely, as it's not common practice to add that for every doc package that includes man pages, and if it was, it'd be done for all recipes, not just this one.

PR #1103, opened 2018-06-13
rng-tools: avoid hardcoding connman

kergoth:
The systemd service exists in oe-core now, we just want it to run
earlier, so adjust it using a drop-in unit rather than overriding the
service file, and also adjust it to run before sysinit.target, rather
than hardcoding connman.

### DIFF
--- a/meta-mentor-staging/recipes-support/rng-tools/rng-tools/10-sysinit.conf
+++ b/meta-mentor-staging/recipes-support/rng-tools/rng-tools/10-sysinit.conf
@@ -1,0 +1,4 @@
+[Unit]
+DefaultDependencies=no
+After=systemd-udev-settle.service
+Before=sysinit.target
--- a/meta-mentor-staging/recipes-support/rng-tools/rng-tools/rngd.service
+++ b/meta-mentor-staging/recipes-support/rng-tools/rng-tools/rngd.service
@@ -1,9 +1,0 @@
-[Unit]
-Description=Hardware RNG Entropy Gatherer Daemon
-Before=connman.service
-
-[Service]
-ExecStart=/usr/sbin/rngd -f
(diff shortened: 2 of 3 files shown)


kergoth:
This is in preparation for upstream submission, as I believe this will be more likely to be accepted.

kergoth:
I'd like to see a bit more testing of this, @Noor-Ahsan, think we can get folks with hardware access to test this on their boards? I could test it with the uno, but I'd like to see it tested on all the boards we're targeting for sumo/elm (though this change won't go in until after that).

Noor-Ahsan:
@kergoth  I think we don't need connman fix. it was issue in the kernel and we fixed this issue for kernel side for couple of BSP. @abelal @fahadarslan  can you confirm.

kergoth:
Ah, good to know, yes, please confirm.

abelal:
@Noor-Ahsan, you're correct. The kernel entropy generator had some issues in the past (4.4 and earlier) which caused problems/lags in various situations. The entropy generator was updated (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/char/random.c?id=e192be9d9a30555aae2ca1dc3aad37cba484cd4a) to cope for such issues. That said, the change here is a different thing and caters the same sort of problem in a different way and IMO is the correct thing to do but we should still test this before merging.

kergoth:
@Noor-Ahsan  , you requested I revisit this, do we need it for sumo/elm, or fir?

kergoth:
It seems this change went into upstream and caused shutdown problems, so a patch was sent to the list to revert it, so need to investigate this issue.

kergoth:
This was fixed in a better way upstream.

PR #1109, opened 2018-07-06
Fixes for newer toolchains

kergoth:
JIRA: SB-11462


kergoth:
Closing, we'll update to newer sumo on oe-core instead.

PR #1110, opened 2018-07-06
Elm: initial toaster support

abelal:
This fixes toaster based builds with the elm installer structure. For details of the patches submitted upstream additionally for this work please see the JIRA ticket below.

JIRA Ticket: INTAMDDET-2413

### DIFF
--- a/scripts/toaster_configuration
+++ b/scripts/toaster_configuration
@@ -218,7 +218,7 @@ if [ "\$#" -ne 1 ]; then
 else
   if [ "\$1" == "start" ]; then
     . $BUILDDIR/venv/bin/activate
-    . $BITBAKEDIR/bin/toaster start
+    . $BITBAKEDIR/bin/toaster start toasterdir="$BUILDDIR"
     xdg-open "http://localhost:8000"
   elif [ "\$1" == "stop" ]; then
     . $BITBAKEDIR/bin/toaster stop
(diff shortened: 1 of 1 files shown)


kergoth:
This looks rather confused, either broken or just confusing and incorrect variable naming. BUILDDIR is where the build is done, aka where conf/local.conf lives, whichi s also TOPDIR. oe-core is relative to MELDIR, aka COREBASE/.., not TOPDIR/BUILDDIR.

abelal:
@kergoth, thanks for the review.
I believe you are referring to https://github.com/MentorEmbedded/meta-mentor/pull/1110/commits/0044e80c1c8022da795cd5ca72732a2a8258ff0a, right?
With the current installer structure that we're following for elm I think MELDIR/BUILDDIR resolves to the same thing every time or else I might have just missed something. Can you please elaborate a little?

kergoth:
That's the default, yes, but there's nothing preventing the user from creating multiple build directories once setup-mel has run. We probably shouldn't assume they're identical unless we can't avoid it.

abelal:
@kergoth, the script we're modifying here gets run while setup-mel is running so it should be safe but I guess in order to make the intent clear we should update oe-core path resolution to use MELDIR here. I'll update the PR.

abelal:
PR has been updated.

PR #1113, opened 2018-07-09
archive-release: make git hashes deterministic for subdirs

kergoth:
The hashes listed in the manifests for layers which are subdirectories and
which are shipped individually were changing on every build. This fixes that
issue.

JIRA: SB-11566


kergoth:
This also makes it fall back to the mtime of layer.conf for layers which aren't git repositories.

PR #1121, opened 2018-07-26
Rework SDK/ADE relocation

kergoth:
Also fix a couple minor build bugs.


meta-mel/conf/distro/include/mel.conf
```diff
@@ -358,7 +358,7 @@ IMAGE_CLASSES += "image-sdk-multilib-variants"
 # We want to be able to run the SDK/ADE relocation script when we first source
 # the environment setup, rather than just at install time. This means the
 # SDK/ADE can be moved, and also works with non-shar installers like the ADE.
-INHERIT += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'mentor-staging', 'sdk_auto_relocate', '', d)}"
+# INHERIT += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'mentor-staging', 'sdk_auto_relocate', '', d)}"
```
Noor-Ahsan: @kergoth isn't it better to remove this line instead of commenting it out if we don't need it?

kergoth: Yeah, will do, forgot I had that :)

kergoth: Fixed, thanks.

PR #1125, opened 2018-08-02
Revert "Merge pull request #1088 from abdulkarim17/SB-11200"

kergoth:
This reverts commit 7bf631d1cabe238b54a2586a35606b5d28b9eb1f, reversing
changes made to 0f0b597d683ee156d191ec55c7bac9f5aed8c1c3.

This change was intended to make /etc/systemd writable on
read-only-rootfs images, but 1) this only makes sense if it's also
persistent, and 2) the motivation was to handle the ability of
opkg-configure to disable itself on first boot, but we don't run
postinsts on first boot on read only images at all, and the case where
we want a persistently writable /etc/systemd is already handled in our
swupdate layer's user-data implementation.

JIRA: SB-8655


kergoth:
Oops, there's one piece left. It seems mounting /etc/systemd for user data is currently in meta-mentor-bsp's append of mentor-swupdate, not the mentor-swupdate layer itself where it belongs. I'll open another pull req for that piece.

kergoth:
The other piece is submitted as a pull request on the other layer, so this is good to go now.

Noor-Ahsan:
@kergoth  can you look at the commit summary again. I think you mentioned wrong commits. https://github.com/MentorEmbedded/meta-mentor/commit/0f0b597d683ee156d191ec55c7bac9f5aed8c1c3 is totally a different commit.

kergoth:
The reference to 0f0b597 was added by git itself when I ran `git revert`. No idea why it picked the wrong one, but I'll remove the incorrect reference.

kergoth:
Fixed, thanks @Noor-Ahsan

PR #1126, opened 2018-08-03
Remove the 'role' image features

kergoth:
- nas
- print-server
- router

JIRA: SB-11707


kergoth:
We're still using tools-audio, I think tools-benchmark may still provide value, and afaict all the rest are pretty clearly still in use, or are the newer image features, so I think this is enough for now.

PR #1128, opened 2018-08-10
Allow gplv3 binutils for aarch64, for tools-profile

kergoth:
The gplv2 binutils does not support aarch64, so the gplv3 version is built and
installed with tools-profile on that architecture by default, even in
a non-gplv3 build. This is kept in local.conf next to the bits which install
binutils with tools-profile by default, to keep it both user-visible and easy
to disable.

Also show a warning when whitelisted incompatible packages are installed in an
image, to ensure the user stays aware of it.

JIRA: SB-11714


kergoth:
The perf fixes are also needed to complete a gplv3 build if you use GPL-3.0 in INCOMPATIBLE_LICENSE rather than GPLv3, so I felt it was worth picking up as well. I'd had those fixes lurking in one of my pending-upstream branches for ages.

PR #1137, opened 2018-09-04
mel: drop sato, fix systemd target for graphics feature

kergoth:
JIRA: SB-11817


sajjad238:
@Noor-Ahsan looks perfect to me.

PR #1138, opened 2018-09-05
Remove new workspace/builddir on failure

kergoth:
JIRA: SB-11880


kergoth:
Only new workspaces/builddirs are removed, to avoid removing existing user content for the reconfiguration use case.

kergoth:
Verified passing `-l invalid` with an existing workspace and builddir does not delete anything, and passing `-l invalid` without an existing workspace or builddir results in the removal of both.

PR #1142, opened 2018-09-07
setup-rh: rework to use pyenv for python3

kergoth:
epel-release's packages cause problems on some rhel/centos systems
depending on the particular versions of the dependencies which are
installed. We have multiple open bugs relating to this. Rework to use pyenv
to install python3.

JIRA: SB-11887


kergoth:
This does have an impact on how long it takes setup-rh to run, since it's building python3 from source, but it avoids issues we've been hitting, so I think it's worth it.

PR #1144, opened 2018-09-17
alsa-lib: drop unnecessary patch

abelal:
The alsa-lib implementation now handles the same situation
more gracefully and after the alsa-lib version upgrade the
patch does not even apply cleanly which additionally shows
warnings at build time.


abelal:
@kergoth, @Noor-Ahsan, please review.

PR #1147, opened 2018-09-19
encrypted-fs image feature for ultrazed

arshadaleem66:
User needs to add "encrypted-fs" to EXTRA_IMAGE_FEATURES in local.conf to enable encryption. Encrypted rootfs can be made through user-space after booting the target with usual rootfs.

Same changes will apply for UNO bsp as well. An additional change in meta-mentor-bsp is made for UNO by adding wks file.


arshadaleem66:
@kergoth the fragment being added by local.conf.sample is not bsp dependent, that's why we added it here instead of bsp layer.

And I'm updating the location of the checks based on EXTRA_IMAGE_FEATURES from mel.conf to local.conf.sample.

PR #1150, opened 2018-09-25
setup-mel/setup-builddir: only setup a workspace

abelal:
The setup-builddir script created a builddir itself if
one wasn't provided by the user and then it would call
setup-environment for it. This causes ambiguities where
a user would want to setup different builds using the
same workspace.
Now we drop the calls to setup-environment and do not
create a builddir from these scripts so the old mechanism
of calling setup-environment separately can be used as
earlier and avoid confusions. The setup-builddir script
is renamed to setup-workspace so the intent is clear.

### DIFF
--- a/scripts/release/setup-mel
+++ b/scripts/release/setup-mel
@@ -10,17 +10,10 @@ else
     fi
     meldir=`readlink -f "$meldir"`
 
-    BUILDDIR=
     WORKSPACEDIR="$PWD/workspace"
     for i in $(seq $#); do
         mel_arg="$(eval printf "%s" "\$$i")"
         case "$mel_arg" in
-            -b)
-                BUILDDIR="$(eval printf "%s" "\$$(expr $i + 1)")"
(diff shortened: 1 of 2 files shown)


abelal:
@Noor-Ahsan, @kergoth, please review this is as per the email conversation we've had around this.

Noor-Ahsan:
looks reasonable to me. Lets wait for @kergoth

abelal:
@kergoth, I understand your point but the thing is we can probably keep this as is and that might be more beneficial.
Say, in future, we decide that the setup script (setup-mel) needs to do some additional stuff and in that scenario we'll be looking at changing the name again so keeping the name generic (i.e. setup-mel) as we have right now might make more sense here and for the same reasons we should probably keep the setup-workspace as we have here to keep a more modular scheme so different functionalities will be implemented using different scripts. What do you think?

Noor-Ahsan:
@kergoth I think let go with this change

PR #1159, opened 2018-10-04
mel: drop INDIVIDUAL_MANIFEST_LAYERS

kergoth:
We aren't using update or addon manifests, drop this.


Noor-Ahsan:
should we remove it from archive-release.bbappend too?

kergoth:
I don't really see the point in that. Keeping the ability to split them doesn't cost much other than a few lines of code, and there's no telling if our requirements will change again in the future.

kergoth:
Fixed

PR #1161, opened 2018-10-05
mel: drop the pointless GLIBC_GENERATE_LOCALES override

kergoth:
The user may well want to have the option to use or install any or
multiple locales, and not generating fr-fr and de-de breaks the ability
to run the bash ptests.

JIRA: SB-11898


Noor-Ahsan:
@kergoth  what is the reason you closed this PR?

kergoth:
I reassigned the original issue back to the person it was assigne dto previously to allow them to do the work if they chose to do so.

PR #1166, opened 2018-10-11
setup-workspace: rework argument handling

kergoth:
Originally, this script was a wrapper script, so used non-standard
argument handling to allow it to pass unrecognized arguments through to
the underlying script directly. Since it was reworked to only set up the
workspace, this is no longer the case, so the old argument handling is
not appropriate. Rework to use getopts and properly show errors/usage on
incorrect arguments.

JIRA: SB-11943

### DIFF
--- a/scripts/release/setup-workspace
+++ b/scripts/release/setup-workspace
@@ -13,25 +13,6 @@ usage () {
     echo >&2
 }
 
-quote () {
-    sed -e "s,','\\\\'',g; 1s,^,',; \$s,\$,',;" << EOF
-$1
-EOF
-}
-
-save () {
-    case "$1" in
-    # when a string contains a "'" we have to escape it
-    *\'*)
-        saved="$saved $(quote "$1")"
-        ;;
-    # otherwise just quote the variable
-    *)
-        saved="$saved '$1'"
-        ;;
-    esac
-}
-
 abspath () {
     _path="$1"
     if [ -n "${_path##/*}" ]; then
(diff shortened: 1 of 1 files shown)


scripts/release/setup-workspace
```diff
@@ -40,79 +21,73 @@ abspath () {
...
+if [ "$1" = "--help" ]; then
+    usage
+    exit 1
+fi
+
 WORKSPACEDIR=$PWD/workspace
-saved=
 manifest=
-extra_disabled=0
+extra_disabled=
```
abelal: @kergoth, this is causing
```
$ source setup-mel -w 123 -x manifests/extra/meta-mentor-industrial/mel-2018.9.29-meta-mentor-industrial.manifest -X
/data/mgc-uno/embedded/mel/2018.9.29/scripts/setup-workspace: line 66: [: : integer expression expected
Error: -x and -X are mutually exclusive
```
pointing out the error seen on line 66 as we use integer test there.

scripts/release/setup-workspace
```diff
@@ -121,11 +96,8 @@ else
 fi
 
 scriptdir="$(dirname "$0")"
-if [ $extra_disabled -eq 1 ]; then
-    "$scriptdir/mel-checkout" -X "$WORKSPACEDIR" "$manifest"
-else
-    "$scriptdir/mel-checkout" "$WORKSPACEDIR" "$manifest" $extra_manifests
-fi && cd "$WORKSPACEDIR"
+"$scriptdir/mel-checkout" ${extra_disabled+-X} "$WORKSPACEDIR" "$manifest" $extra_manifests \
```
abelal: @kergoth, it looks like the -x option is not behaving as it should and although extra manifests are being passed to mel-checkout the -X option always shows up and borks the execution.

scripts/release/setup-workspace
kergoth: Ah, missed one. Thanks, I'll fix it.

scripts/release/setup-workspace
kergoth: Hmm, interesting. ${foo+bar} expands to bar only when foo is set, generally. I'll take a closer look, thanks for testing. I did do basic build testing with this and it worked fine there, but I must have missed something.

kergoth:
@abelal, please re-test, both issues should be fixed now.

kergoth:
@abelal there are manifests which are included implicitly. Specifically, any update/addon manifests are automatically included when available. -X allows you to prevent this.

PR #1167, opened 2018-10-16
MEL Lite use glibc 2.28

abelal:
The CB Lite tools for current release are based off of
glibc 2.28 so make adjustments accordingly.


kergoth:
Hmm, my only thought is whether this truly belongs in meta-mentor. Yes, it's mel-lite, but even so, this would be easy to lose track of amongst everything else in meta-mentor. Thoughts?

abelal:
@kergoth, we can probably create a Jira ticket to track reversal of this in the next release? Apart from that this looked to me as the right place for this update.

Noor-Ahsan:
lets go with this change and we open an issue to track future changes.

PR #1170, opened 2018-10-20
Add support for ROOT_PASSWORD, change behavior when empty for production-image

kergoth:
WIP, do not merge yet. I don't believe this is a good approach, defaulting to
an insecure random password for root in production-image is problematic. I'd
rather arrange for a warning to be emitted when production-image is built when
the user hasn't set ROOT_PASSWORD, indicating that they won't be able to login
as root by default.


kergoth:
Please add your thoughts, I don't think this is an ideal approach. This is a fixed version of @fahadarslan  's original implementation currently.

chombourger:
I'm fine if we force the user to set a password, or use "0" to allow the empty password or to use e.g. "*" to disable root login (in which case, shell for the root user could be set to /bin/nologin)

kergoth:
Thanks @chombourger, I'll rework the pull request.

chombourger:
Perfect, thanks!

PR #1180, opened 2018-10-31
python-native: fix one do_populate_sysroot warning

kergoth:
Fix below warning:
WARNING: Skipping RPATH /usr/lib64 as is a standard search path for
work/x86_64-linux/python-native/2.7.15-r1.1/recipe-sysroot-native/
usr/lib/python2.7/lib-dynload/_bsddb.so

setup.py will check db.h under include_dirs, for native build,
/usr/lib64 will be insert to postion 0 of include_dirs, so
it's priority is higher then our sysroot, cause db.h sysroot
is ignored, and rpath set to /usr/lib64. and this cause warning
when do_populate_sysroot. use append to fix it.

Signed-off-by: Changqing Li <changqing.li@windriver.com>


kergoth:
This fixes SB-12077.

PR #1181, opened 2018-10-31
mel.conf: update supported distros for elm

kergoth:
- Removed ubuntu-14.04
- Also match centos-7, not just centos-7.*
- Also match rhel, not just redhatenterprise

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -91,10 +91,10 @@ PACKAGE_CLASSES ?= "package_ipk"
 SANITY_TESTED_DISTROS = "\
     ubuntu-18.04 \n\
     ubuntu-16.04 \n\
-    ubuntu-14.04 \n\
-    centos-7.* \n \
-    centoslinux-7.* \n \
-    redhatenterprise*-7.* \n \
+    centos-7* \n \
+    centoslinux-7* \n \
+    rhel*-7* \n \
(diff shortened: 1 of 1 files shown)


Noor-Ahsan:
@kergoth  should we be specific to the versions we are supporting. See MR http://jira.alm.mentorg.com:8080/browse/MELMR-731

kergoth:
Ah, indeed, I didn't realize rhel was limited to 7.3 specifically. I'll update the pull req.

mwpow3ll:
@kergoth @Noor-Ahsan Just a random note here... CentOS builds from the same sources with very minimal changes; therefore, if support is limited to RHEL 7.3, shouldn't CentOS be limited to 7.3.1611?

kergoth:
@mwpow3ll , it's a good question, wasn't my call, would you mind adding your thoughts to the MELMR so we can update it there before changing it here? Thanks.

chombourger:
Shouldn't we support RedHat 7.x instead of specific minor version(s)?

kergoth:
I'm pretty sure there can be more substantial changes between rhel/centos minor versions, enough to potentially break our host setup scripts, but I'm not 100% certain.

Noor-Ahsan:
Kindly update this PR according to comments in http://jira.alm.mentorg.com:8080/browse/MELMR-731

kergoth:
@Noor-Ahsan, ready for merge.

PR #1183, opened 2018-11-02
staging/layer.conf: hack around LAYERSERIES_COMPAT for qt5-layer

abelal:
Adding to the tradition set in 3c373331, fix QA warning
WARNING: Layer qt5-layer should set LAYERSERIES_COMPAT_qt5-layer in its conf/layer.conf file to list the core layer names it is compatible with.


Noor-Ahsan:
@abelal  instead of adding me in reviewers kindly assign the PR to me. That is more helpful for me in case I miss reviewer email.

PR #1184, opened 2018-11-05
setup-mel-builddir: set BBLAYERS using loose assignment

abelal:
In cases such as toaster the setting for BBLAYERS is
done through cmdline params and then during the parse
phase bblayers.conf is picked up which overrides the
setting that was passed if we use a simple set (=)
here.
Fix this by using loose assignment so cmdline params
work as expected.

JIRA: SB-12082.


abelal:
Although I think toaster should try to implement things in a way that allows overriding everything, this is still correct. Discussion about how toaster should work itself is going on with David Reyna, the toaster maintainer.
One more thing to note here is oe's scripts also use a loose assignment for BBLAYERS in the bblayers.conf that it generates: http://cgit.openembedded.org/openembedded-core/tree/meta/conf/bblayers.conf.sample?h=sumo

Noor-Ahsan:
@kergoth kindly review this.

PR #1185, opened 2018-11-05
mel-image.inc: Add haveged as entropy source.

arshadaleem66:
If GPLv3 license is allowed to be used, haveged is added as an entropy
source so that if user does not use random input sources like keyboard
and mouse, MEL image still has a random number's source.

### DIFF
--- a/meta-mel-support/recipes-core/images/mel-image.inc
+++ b/meta-mel-support/recipes-core/images/mel-image.inc
@@ -10,4 +10,4 @@ LICENSE = "MIT"
 inherit core-image image-sanity-incompatible
 
 IMAGE_FEATURES .= "${@bb.utils.contains('COMBINED_FEATURES', 'alsa', ' tools-audio', '', d)}"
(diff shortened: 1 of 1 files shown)


Noor-Ahsan:
@arshadaleem66 have you validated docker support with this?

arshadaleem66:
@Noor-Ahsan yes I have validated it for docker support on ZCU102

arshadaleem66:
@abelal, yes I have documented it here: http://confluence.alm.mentorg.com:8090/display/~aaleem/Latency+due+to+Random+Number+Generator%27s+Unavailability
And acknowledged about the Jira ticket

kergoth:
This should be using the function in meta-mentor which handles license WHITELISTing, rather than directly using bb.utils.contains(). also, contains() won't handle GPL-3.0 being in INCOMPATIBLE_LICENSE, so that's another problem with it.

kergoth:
@arshadaleem66, please submit a pull req to change that.

PR #1189, opened 2018-11-07
update_font_cache: fix warnings

kergoth:
This series applies a couple patches from oe-core which fixes the
update_font_cache failures seen on x86-64 and elsewhere relating to multilibs,
adjusted to work in meta-mentor-staging without overriding the classes in
question. We no longer see warnings for this postinst intercept with this
applied.

JIRA: SB-12123


Noor-Ahsan:
@kergoth  as you brought it from upstream. what will happend when we will update the oe-core layer for bug fixing. Will we have to remove it at the time?

kergoth:
> @kergoth as you brought it from upstream. what will happen when we will update the oe-core layer for bug fixing. Will we have to remove it at the time?

@Noor-Ahsan, No, these are fixed in master, not sumo.

PR #1194, opened 2018-11-09
dmcrypt-initramfs-image: use INITRAMFS_FSTYPES

kergoth:


kergoth:
This is needed for uno for sb-12130.

PR #1199, opened 2018-11-15
busybox: enable reboot support

abelal:
996247ba7dfffbeb444f793f7e105fcfb5ffa939 in upstream oe-core
moved the power flags (POWEROFF, HALT, REBOOT) to a fragment
init.cfg which is only included when busybox is selected as
the init_manager. This assumption is incorrect in scenarios
where installation mechanism is provided through live images
and almost all the initramfs scripts use the reboot command at
the end of installation. Without this change we see
reboot: not found


abelal:
@Noor-Ahsan, please review.

abelal:
@Noor-Ahsan, thanks!

PR #1200, opened 2018-11-16
connman: Restrict connman from including it's DNS proxy

abdulkarim17:
with virtualization-layer dnsmasq comes,so we do not use connman as
a DNs proxy because both dnsmasq and connman try to bind to same
port 53. which results in failure of dnsmasq service.

Jira: http://jira.alm.mentorg.com:8080/browse/SB-12022


kergoth:
This sounds like it'll break or at the least lose functionality if dnsmasq isn't included for non-virtualization builds? @abdulkarim17 ?

PR #1203, opened 2018-11-22
Kill Volatile Files and Directories Failure status

arshadaleem66:
JIRA: SB-12126


meta-mentor-staging/recipes-connectivity/connman/connman_%.bbappend
```diff
@@ -1,5 +1,6 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
-SRC_URI += "file://0001-connman-implement-network-interface-management-techn.patch"
+SRC_URI += "file://0001-connman-implement-network-interface-management-techn.patch \
+            file://fix-resolv-link.patch"
```
Noor-Ahsan: what is the behavior of this patch when there is not readonly fs?

kergoth: It's a non-issue for us, as the recipe already creates this symlink directly.

PR #1204, opened 2018-11-22
toaster_configuration: conform to new mechanism of BBLAYERS

abelal:
BBLAYERS is now set using a weak assignment rather than a
strict one. See 6ac988dcad81b6bfa38e15446618184258d23eba
in meta-mentor. Without this the BBLAYERS variable isn't
parsed and picked up correctly.

JIRA: SB-12165.

### DIFF
--- a/scripts/toaster_configuration
+++ b/scripts/toaster_configuration
@@ -18,7 +18,7 @@ OE_CORE_NAME="openembedded-core"
 BBLAYERS=
 configured_layers () {
    tac "$BUILDDIR/conf/bblayers.conf" | \
-   sed -n -e '/^"/,/^BBLAYERS = /{ /^BBLAYERS =/d; /^"/d; p;}' | \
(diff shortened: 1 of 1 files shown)


kergoth:
This should ideally be more flexible, handling both cases, rather than hardcoding the assumption that it's always ?=, to be more future-proof.

abelal:
@kergoth, makes sense. How about
`sed -n -e '/^BBLAYERS/,/^"/{/^BBLAYERS/d;/^"/d;p;}' "$BUILDDIR/conf/bblayers.conf" | tac | awk {'print $1'}`

I inclined towards this because it does not rely on a line starting with " for starting up the match as that assumption can go haywire with a small change. Also this keeps the other assumption that we want to print the lines backwards which might be for a different reason but lets not dig into that for now as the whole script clearly requires some cleanup and can be optimized to a better shape for sure. Probably that can be a task for the next release.

Noor-Ahsan:
@abelal we can update this in async. This will be picked as an update to main elm release. let's hold it for a week or so.

PR #1207, opened 2018-11-26
SB-12080

abelal:
Minor cleanup and fix for SB-12080.

### DIFF
--- a/scripts/toaster_configuration
+++ b/scripts/toaster_configuration
@@ -224,6 +224,7 @@ else
   elif [ "\$1" == "stop" ]; then
     . $BITBAKEDIR/bin/toaster stop
     deactivate
+    unset CUSTOM_XML_ONLY
   else
      echo >&2 "Usage: . ./toaster-setup-environment {start|stop}"
   fi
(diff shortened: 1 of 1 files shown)


abelal:
@Noor-Ahsan, the upstream bitbake 1.38 now has the other part for this fix available: http://cgit.openembedded.org/bitbake/log/?h=1.38.

scripts/toaster_configuration
```diff
@@ -218,13 +218,17 @@ if [ "\$#" -ne 1 ]; then
 else
   if [ "\$1" == "start" ]; then
     export CUSTOM_XML_ONLY=1
+    export OE_ROOT="$POKYDIR"
```
Noor-Ahsan: We are not using poky distro now. Is it wise to rename this variable?

abelal: @Noor-Ahsan, yes that would be wise but then the whole script requires some attention :) because there are various parts that can be overhauled.

Noor-Ahsan: Yeah but we removed all poky related varaibles from meta-mentor. I was thinking atleast we update poky related var in this script too.

abelal: @Noor-Ahsan, agreed. I'll push the update shortly.

Noor-Ahsan: thanks

abelal:
@Noor-Ahsan, updated as discussed, please review.

PR #1210, opened 2018-11-30
Exclude vmlinux from production-image

kergoth:
We'll need to discuss whether this is appropriate for sumo for patch releases and asyncs or if it should go directly to master. There's also the question of whether to default to exclusion for all images and explicitly include it in development-image, or this approach, which leaves it included in images by default and excludes it explicitly in production-image.

### DIFF
--- a/meta-mel-support/recipes-core/images/production-image.bb
+++ b/meta-mel-support/recipes-core/images/production-image.bb
@@ -12,3 +12,7 @@ IMAGE_FEATURES_DISABLED_PRODUCTION ?= "debug-tweaks codebench-debug tools-profil
 IMAGE_FEATURES_remove = "${IMAGE_FEATURES_DISABLED_PRODUCTION}"
 
 require recipes-core/images/mel-image.inc
+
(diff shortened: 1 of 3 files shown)


abelal:
@kergoth, the implementation here is correct but I believe there's more to the issue we're trying to address. As per my discussions locally with @ArsalanHAwan who was also looking into this it turns out that the vmlinux binary is generated by the kernel build process no matter what (i.e. vmlinux is not required in KERNEL_IMAGETYPES) but is only deployed when the KERNEL_IMAGETYPES variable contains vmlinux. That said, the kernel debugging flow documented right now for MEL also asks the user to search for the vmlinux binary inside the kernel's builddir and that workflow is also well tested.
@ArsalanHAwan, is going to open a new PR with his investigation populated so lets see that as well before making a final decision here.

ArsalanHAwan:
@abelal, @kergoth, please take a look at this: https://github.com/MentorEmbedded/meta-mentor/pull/1211
Thanks!

kergoth:
Fair enough, thanks for the review and your work on this.

PR #1215, opened 2018-12-07
archive-release: don't individually archive non-topdir indiv layers in git repos

kergoth:
This case now archives the git repo as a whole, but filters its contents to
only include the indiv layers in the repo, no other content.

### DIFF
--- a/meta-mel/mel-support/recipes-core/meta/archive-release.bbappend
+++ b/meta-mel/mel-support/recipes-core/meta/archive-release.bbappend
@@ -61,7 +61,22 @@ def mel_get_remotes(subdir, d):
 
 GET_REMOTES_HOOK_mel ?= "mel_get_remotes"
 
+GIT_ROOT_TOO_FAR_PATHS = "${OEDIR} ${TOPDIR} ${HOME}"
+
+def layer_git_root(subdir, too_far_paths):
+    try:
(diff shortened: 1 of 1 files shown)


kergoth:
The effect of this is to end up with `meta-mentor-bsp` being a git repository that only contains common and the appropriate bsp layer(s), rather than a number of individual git repositories for each of the sublayers within meta-mentor-bsp. This is cleaner, closer to how we do development, allows al engineers to use `git fetch —unshallow` on the layer, and still works fine with our git delivery.

kergoth:
Thoughts?

kergoth:
Updated to fix the too far logic in layer_git_root to handle the case where the root is a too far path rather than just being its parent.

kergoth:
This is master, not sumo, as it changes how the layers are archived and invalidates testing, not to mention the code freeze.

Noor-Ahsan:
How this will behave on this https://github.com/MentorEmbedded/mel-meta-mentor-security

kergoth:
> How this will behave on this https://github.com/MentorEmbedded/mel-meta-mentor-security

No different structurally than how we do it today,  the only difference is where the .git is / which dir is a repository. This only affects layers marked as individual only, but not individual only toplevel (as if the layer is moved to toplevel, it should have its own repo).

PR #1216, opened 2018-12-14
Fix mel-lite and codebench-lite handling in setup scripts

abelal:
The workspace setup script needs to pass on information regarding mel-lite variant to the workspace itself so once the scripts from the workspace are executed they can detect the presence of the distro. Furthermore, the setup scripts inside the workspace are fixed to handle the mel-lite and codebench-lite cases.

### DIFF
--- a/scripts/release/setup-workspace
+++ b/scripts/release/setup-workspace
@@ -105,3 +105,6 @@ if [ $? -ne 0 ]; then
     fi
     exit 1
 fi
+if [ -e "$scriptdir/../.mel-lite" ]; then
+    touch "$WORKSPACEDIR/.mel-lite"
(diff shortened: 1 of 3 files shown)


Noor-Ahsan:
Merged for the time being so that AMD DET team can generate installer

kergoth:
For fir I'd really like to see this improved. There's no real need to jump through these hoops given the setup scripts already provide hooks to modify things. All you really need for mel-lite is a mel-lite layer that has a post-setup-environment hook to correct CODEBENCH_PATH/EXTERNAL_TOOLCHAIN and a local.conf.append to override the default DISTRO, as far as I can tell.

PR #1227, opened 2019-02-27
local.conf: replace += with append_ operator

arshadaleem66:
IMAGE_BOOT_FILES is being assigned in machine file and being appended
in local.conf.
Bitbake parses local.conf earlier and parses machine file later.
Also, += operator appends the corresponding variable instantly during
the parse.
And _append operator appends the variable after parsing all the recipes
completely.
In case we place += operator in local.conf and same variable is being
assigned (using = operator) in other recipe, we will get our local.conf
append overwritten due to later assignment. This overwrite is simply
avoided by using _append instead of += operator.

Jira Issue: http://jira.alm.mentorg.com/browse/SB-12629

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -303,7 +303,7 @@ BB_DISKMON_DIRS = "\
 # removing 96boards-tools package so that rootfs does not occupy entire available space over the boot media.
 MACHINE_EXTRA_RRECOMMENDS_remove = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', '96boards-tools', '', d)}"
 # Add initramfs image to the boot partition.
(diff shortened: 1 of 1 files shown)


kergoth:
@arshadaleem66  , unless dm-initramfs already includes a leading space, this is broken. `_append` doesn't add a leading space for you the way `+=` does.

arshadaleem66:
> @arshadaleem66 , unless dm-initramfs already includes a leading space, this is broken. `_append` doesn't add a leading space for you the way `+=` does.

@kergoth yes, I got it. Leading space is not being added by dm-initramfs. Probably it didn't break while testing due to 'bad' luck factor of some spaces already being added into IMAGE_BOOT_FILES. Let me fix it.

PR #1229, opened 2019-03-14
Encrypted-fs support updates.

arshadaleem66:
**encryptfs.bbclass: conditionally add fragment** 
kernel-config-encryptfs.bbclass adds kerenl configs required by
encrypted file system support. And this class is inherited in
kernel recipe class. This class adds the kernel configs in all
cases, even if encrypted-fs is not enabled as a feature.

This commit adds this kernel fragment to SRC_URI only if the
encrypted-fs is enabled in the USER_FEATURE variable.
Otherwise doesn't add it.

**local.conf: set encrypted-fs as a user feature**
encrypted-fs was previously being used as an EXTRA_IMAGE_FEATURE.
But machine related kernel configurations cannot be made on the
basis of an image feature. Since in most of the imx bsps, we
encountered hardware crypto failure, kernel configurations needed
to be set on the basis of encrypted-fs. And that's why it is
to be set as USER FEATURE instead of IMAGE FEATURE.

This commit updates the checks that were previously being made
on the basis of IMAGE FEATURE and lets the user to add encrypted-fs
as USER_FEATURE just by uncommenting a line.

**mel.conf: move encryptedfs configuration to distro**
Encrypted File System support needs to remove 96boardstools and
add kernel image to boot partition.  Both of these actions were
previously being performed on the basis of a check condition
applied to EXTRA_IMAGE_FEATURE. And due to intervention of the
IMAGE_FEATURE variable, these statements were placed in local.conf

But now, as we switched encrypted-fs to USER_FEATURE, and there
is no user intervention intended with these statements, we can now
move them to mel.conf.

Additionally, DM_INITRAMFS variable was previously being set by
each bsp in their machine file. The variale stored name of the
bundled image of initramfs+kernel so that it could be placed in
boot partition via IMAGE_BOOT_FILES variable. But same action can
be performed by yocto provided variables, as the name of bundled
file keeps same pattern. This change is also added to mel.conf
along the being shifted configuration lines for encrypted-fs.

Jira Issue: http://jira.alm.mentorg.com:8080/browse/SB-12527

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -44,9 +44,6 @@ MACHINE ??= "qemux86"
 #
 #  "virtualization"  - add docker support. Including meta-virtualization and dependencies
 #                      to bblayers is COMPULSORY for having no errors.
-#
-#  "encrypted-fs"     - add encrypted file system support. Some manual steps are needed
-#                       after flashing the image.
 # There are other options that can be used here, see
 # meta/classes/image.bbclass and meta/classes/core-image.bbclass for more
 # details.
@@ -296,14 +293,17 @@ BB_DISKMON_DIRS = "\
 # Use this to add extra packages to the split-debug filesystem
 #EXTRA_DEBUG_PACKAGES = "ltrace strace"
 
-# Uncomment following two lines to enable INITRAMFS. This adds dmcrypt-initramfs-image in 
-# INITRAMFS_IMAGE if encrypted-fs image feature is enabled otherwise add core-image-minimal-initramfs.
-#INITRAMFS_IMAGE = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', 'dmcrypt-initramfs-image', 'core-image-minimal-initramfs', d)}"
+# Uncomment to enable encrypted-fs User Feature:
+#USER_FEATURES_append = " encrypted-fs"
+EXTRA_IMAGE_FEATURES_append = " ${@bb.utils.contains('USER_FEATURES', 'encrypted-fs', 'encrypted-fs', '', d)}"
+# Uncomment it to enable INITRAMFS:
+#INITRAMFS_IMAGE = "${@bb.utils.contains('USER_FEATURES', 'encrypted-fs', 'dmcrypt-initramfs-image', 'core-image-minimal-initramfs', d)}"
+# Uncomment it to bundle initramfs image with kernel image:
 #INITRAMFS_IMAGE_BUNDLE = "1"
 # removing 96boards-tools package so that rootfs does not occupy entire available space over the boot media.
-MACHINE_EXTRA_RRECOMMENDS_remove = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', '96boards-tools', '', d)}"
+MACHINE_EXTRA_RRECOMMENDS_remove = "${@bb.utils.contains('USER_FEATURES', 'encrypted-fs', '96boards-tools', '', d)}"
 # Add initramfs image to the boot partition.
-IMAGE_BOOT_FILES_append = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', '${dm-initramfs}', '', d)}"
+IMAGE_BOOT_FILES_append = " ${@bb.utils.contains('USER_FEATURES', 'encrypted-fs', '${dm-initramfs}', '', d)}"
 
 # CONF_VERSION is increased each time build/conf/ changes incompatibly and is used to
 # track the version of this file when it was generated. This can safely be ignored if
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -300,10 +300,6 @@ EXTRA_IMAGE_FEATURES_append = " ${@bb.utils.contains('USER_FEATURES', 'encrypted
 #INITRAMFS_IMAGE = "${@bb.utils.contains('USER_FEATURES', 'encrypted-fs', 'dmcrypt-initramfs-image', 'core-image-minimal-initramfs', d)}"
 # Uncomment it to bundle initramfs image with kernel image:
 #INITRAMFS_IMAGE_BUNDLE = "1"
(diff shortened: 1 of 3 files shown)


conf/local.conf.sample
```diff
@@ -296,14 +293,16 @@ BB_DISKMON_DIRS = "\
...
-#INITRAMFS_IMAGE = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', 'dmcrypt-initramfs-image', 'core-image-minimal-initramfs', d)}"
+# Uncomment to enable encrypted-fs Machine Feature:
+#MACHINE_FEATURES_append = " encrypted-fs"
+# Uncomment it to enable INITRAMFS:
+#INITRAMFS_IMAGE = "${@bb.utils.contains('MACHINE_FEATURES', 'encrypted-fs', 'dmcrypt-initramfs-image', 'core-image-minimal-initramfs', d)}"
+# Uncomment it to bundle initramfs image with kernel image:
 #INITRAMFS_IMAGE_BUNDLE = "1"
 # removing 96boards-tools package so that rootfs does not occupy entire available space over the boot media.
-MACHINE_EXTRA_RRECOMMENDS_remove = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', '96boards-tools', '', d)}"
+MACHINE_EXTRA_RRECOMMENDS_remove = "${@bb.utils.contains('MACHINE_FEATURES', 'encrypted-fs', '96boards-tools', '', d)}"
```
abelal: Why aren't we doing this through mel.conf? This is not something we're allowing the user to control I guess.

conf/local.conf.sample
```diff
@@ -296,14 +293,16 @@ BB_DISKMON_DIRS = "\
...
 # Add initramfs image to the boot partition.
-IMAGE_BOOT_FILES_append = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', '${dm-initramfs}', '', d)}"
+IMAGE_BOOT_FILES_append = " ${@bb.utils.contains('MACHINE_FEATURES', 'encrypted-fs', '${dm-initramfs}', '', d)}"
```
abelal: Again, how about doing this through mel.conf? Also please define
`dm-initramfs ?= ""`
somewhere in the configuration so the build wouldn't break for BSPs not defining the variable as investigated after a lot of effort :smile:

arshadaleem66: Yes @abelal, I'd have a discussion with @Noor-Ahsan hopefully today regarding moving these lines to mel.conf.
Secondly, I had a discussion with @ArsalanHAwan. We may drop dm-initramfs in the improvements to come in a few days. I am looking into a composite of yocto's variables that can do the same job as dm-initramfs is doing. Composite will be like this: 
${KERNEL_IMAGETYPE}-${INITRAMFS_NAME}
But I've to take care of some factors, as this change may affect the base release bsps (uno, zcu102, ultrazed) whose documentation is already done according to the flow brought in by dm-initramfs.

conf/local.conf.sample
arshadaleem66: @abelal, it was due to IMAGE_FEATURE thing. As we cannot make decisions based on IMAGE_FEATURES in a non-image recipe. Because user may add/remove an image feature.
But now, as we have moved on to MACHINE_FEATURE implementation of encrypted-fs, it can be negotiated to move these statements to somewhere else.

abelal: mel.conf is not a recipe :)

Noor-Ahsan:
I might need to hold this change for a week. my async2 is at CF. only critical bugs will go. Once it will freeze. I'll merge this and any other changes on meta-mentor.

conf/local.conf.sample
arshadaleem66: * non-image recipe and non-local.conf conf file. Sorry I missed. :)

arshadaleem66:
> I might need to hold this change for a week. my async2 is at CF. only critical bugs will go. Once it will freeze. I'll merge this and any other changes on meta-mentor.

Sure @Noor-Ahsan.
As per our discussion, I have updated this PR with the commit that moves encrypted-fs related contents from local.conf to mel.conf that user is not supposed to change.

conf/local.conf.sample
```diff
@@ -296,14 +293,12 @@ BB_DISKMON_DIRS = "\
...
-# INITRAMFS_IMAGE if encrypted-fs image feature is enabled otherwise add core-image-minimal-initramfs.
-#INITRAMFS_IMAGE = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', 'dmcrypt-initramfs-image', 'core-image-minimal-initramfs', d)}"
+# Uncomment to enable encrypted-fs Machine Feature:
+#MACHINE_FEATURES_append = " encrypted-fs"
+# Uncomment it to enable INITRAMFS:
+#INITRAMFS_IMAGE = "${@bb.utils.contains('MACHINE_FEATURES', 'encrypted-fs', 'dmcrypt-initramfs-image', 'core-image-minimal-initramfs', d)}"
+# Uncomment it to bundle initramfs image with kernel image:
 #INITRAMFS_IMAGE_BUNDLE = "1"
-# removing 96boards-tools package so that rootfs does not occupy entire available space over the boot media.
-MACHINE_EXTRA_RRECOMMENDS_remove = "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'encrypted-fs', '96boards-tools', '', d)}"
```
arshadaleem66: @abelal these lines moved to mel.conf now.

abelal:
@Noor-Ahsan, ping! AMD builds are down.

arshadaleem66:
@abelal yes, I have successfully build tested the changeset. Thanks.

kergoth:
USER_FEATURES is really just a wrapper around DISTRO_FEATURES for easier use in local.conf. There's no reason we couldn't use distro features directly in mel.conf, modify the conditionals to be based on distro features, etc. Of course, altering kernel configuration based on a distro feature is a little hokey, this could po tentially be a combination of both machine and distro features the way we do swupdate in the future. This is fine for now, though.

PR #1230, opened 2019-03-20
setup-ubuntu: install realpath

abelal:
The utility is used by the Host Security Tool's command line
variant.


kergoth:
It really can't function without this? `readlink -f` and even `pwd -P` when dealing with existing directories are viable alternatives.

abelal:
@kergoth, I'll see what CB guys think about that :) thanks for the pointer.

Noor-Ahsan:
We also need to check RHEL and CentOS hosts.

PR #1232, opened 2019-04-08
python3-docker-compose: add missing dependencies

arshadaleem66:
docker-compose needs a number of python3 runtime dependencies. The
packages python3-misc and python3-modules provide those dependencies.
The bbappend is being applied to current version of python3-docker-
compose so that it is reviewed later on as well if these dependencies
are fixed by upstream in the later versions or not.


kergoth:
Looks good, thanks. You'd be amazed how many recipes are missing required python runtime dependencies like this. Back in 2015 I ran a script to collect them -- https://gist.github.com/kergoth/7d2e9f6c7d0fd759d7f2#file-check-python-deps-txt -- I'm sure it's improved since then, but yikes, eh?

PR #1233, opened 2019-04-09
Add kconfig fragments through bbclasses

arshadaleem66:
**Commit 1: kernel-config-flatpak: add namespaces conditionally**
If flatpak-layer is included in build, kernel fragment for namespaces
will be applied. This class can be inherited in any kernel recipe
where flatpak is to be supported.

JIRA issue: http://jira.alm.mentorg.com:8080/browse/SB-12643

**Commit 2: kernel-config-virtualization: add kconfigs in bbclass**
Generic kernel configs for virtualization were previously being
added to each bsp's layer separately to add them into kerenl src
uri.

This commit adds those fragments to SRC_URI inside a class in common
place and that class can be inherited in each bsp's kernel recipe.

The addition of fragments into SRC_URI is performed only when the
virtualization layer is added in build. That's why, this class can
be safely inherited into main kernel recipes. Fragments will only
apply if virtualization is enables.


ahsanhussain:
@arshadaleem66 I think you can squash your two commits, as both are part of a single implementation.

Also update the PR message, as it's part of the git history.

arshadaleem66:
@ahsanhussain sure
@abelal no, not all of them are directly dependent on namespaces. Some (UTS_NS, IPS_NS etc) are dependent on NAMESPACES. I have added all of them because most of them are disabled by default.

Noor-Ahsan:
@arshadaleem66 I think @kergoth  PR contained configs for lxc but lxc configs are present in https://github.com/MentorEmbedded/meta-mentor-bsp/pull/435. So I think this cfg has some extra bits which are not needed. So we need some cleanup a bit

kergoth:
> @arshadaleem66 I think @kergoth PR contained configs for lxc but lxc configs are present in https://github.com/MentorEmbedded/meta-mentor-bsp/tree/elm_async2/imx/common-imx/virtualization-layer/recipes-kernel/linux. So I think this cfg has some extra bits which are not needed. So we need some cleanup to do. @kergoth can you help here as we got htis cfg from your PR. If you can tell what we can drop from this cfg after looking at the link I shared.

My PR was actually a subset of the lxc.cfg we're already pulling in when meta-virtualization is included in meta-mentor-bsp. Any extra configs to support containers aren't going to harm anything, even if they aren't strictlyr equired for bubblewrap alone. Arguably all our kernels should enable all the sane container and namespace options across the board, regardless of whether lxc, bubblewrap, or docker require them, as it's a baseline linux feature now.

arshadaleem66:
@abelal, as discussed, in the new commit I have placed the virtualization related fragments and added them to SRC_URI via a bbclass. Kindly have a look.

meta-mentor-common/classes/kernel-config-flatpak.bbclass
```diff
@@ -0,0 +1,9 @@
+FILESEXTRAPATHS_append = ":${@os.path.dirname(bb.utils.which(BBPATH, 'files/namespaces.cfg') or '')}"
+
+NAMESPACES_CFG = "${@bb.utils.contains('BBLAYERS', 'flatpak-layer', 'file://namespaces.cfg', '', d)}"
```
abelal: Why in one class we're doing this and in the other one we're adding cfg files directly to the SRC_URI?

arshadaleem66: @abelal, no specific reason for it. To conserve uniformity, I'll update the other bbclass like this one.

meta-mentor-common/classes/kernel-config-virtualization.bbclass
```diff
@@ -0,0 +1,13 @@
...
+SRC_URI_append = "${@bb.utils.contains('BBLAYERS', 'virtualization-layer', ' file://docker.cfg \
+                                                                             file://ebtables.cfg \
+                                                                             file://lxc.cfg \
+                                                                             file://vswitch.cfg \
+                                                                             file://xt-checksum.cfg \
+                                                                             ', '', d)}"
+
+python () {
+    if not oe.utils.inherits(d, 'kernel-yocto', 'cml1-config'):
+        bb.fatal("kernel-config-flatpak.bbclass requires kernel config fragments support. Please use kernel-yocto.bbclass or cml1-config.bbclass.")
```
ahsanhussain: s/flatpak/virtualization

meta-mentor-common/classes/kernel-config-flatpak.bbclass
ahsanhussain: +1

meta-mentor-common/classes/kernel-config-virtualization.bbclass
arshadaleem66: Oh, terribly missed it! Thanks.

PR #1240, opened 2019-05-22
add psplash to IMAGE_FEATURES

abdulkarim17:


meta-mel-support/recipes-core/images/development-image.bb
```diff
@@ -5,5 +5,6 @@ IMAGE_FEATURES = "${IMAGE_FEATURES_DEVELOPMENT} ${EXTRA_IMAGE_FEATURES}"
 IMAGE_FEATURES_DEVELOPMENT ?= "debug-tweaks tools-profile"
 IMAGE_FEATURES_DISABLED_DEVELOPMENT ?= ""
 IMAGE_FEATURES_remove = "${IMAGE_FEATURES_DISABLED_DEVELOPMENT}"
+IMAGE_FEATURES += "splash"
```
abelal: As per @Noor-Ahsan's comment on IRC this is needed for production image as well so the best place would be mel-image.inc.

abdulkarim17: @abelal  added in mel-image.inc 
Thanks

PR #1241, opened 2019-05-22
Avoid conflicts between different images for the ade

kergoth:
This is needed to fix the ability to install both `development-image` and `lib32-development-image` ades into CodeBench.

JIRA: SB-12981


meta-mel/conf/distro/include/mel.conf
```diff
@@ -392,7 +392,7 @@ ADE_PROVIDER ?= "Mentor Graphics Corporation"
 OE_IMPORTS += "string"
 ADE_VERSION ?= "${@'0.0' if not all(c in string.digits for c in '${SDK_VERSION}'.split('.')[0]) else '${SDK_VERSION}'}.${DATETIME_SECS}"
 ADE_VERSION[vardepsexclude] += "DATETIME_SECS"
-ADE_IDENTIFIER ?= "${SDKMACHINE}-${MLPREFIX}${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}"
+ADE_IDENTIFIER ?= "${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${ADE_VERSION}"
```
Noor-Ahsan: What will happen if we don't revert it for non multilib ades. this got included in Patch release. will it has any impact?

kergoth: The only impact is cosmetic, and even that won't be visible unless they disable the category view when installing, and that will only be different when in a multilib configuration.. It just shows 'lib32-lib32-development-image' instead of 'lib32-development-image'.

PR #1242, opened 2019-06-28
openssh: drop sshd_config override

lumag:
sshd_config override lists only deprecated KexAlgorithms resulting in
unability to connect to boards with contemporary ssh using default
parameters. These algorithms are considered weak and thus should not be
enabled in the product.

Signed-off-by: Dmitry Eremin-Solenikov <dmitry_eremin-solenikov@mentor.com>


Noor-Ahsan:
Do you want this for elm release. if yes then you have to submit PR for sumo branch.

lumag:
@Noor-Ahsan Strictly speaking I do not  care, as I used local override. I'll open one.

PR #1244, opened 2019-07-09
mel-image: add psplash to IMAGE_FEATURES in graphical images only

abdulkarim17:
Jira: http://jira.alm.mentorg.com:8080/browse/SB-13336

### DIFF
--- a/meta-mel-support/recipes-core/images/mel-image.inc
+++ b/meta-mel-support/recipes-core/images/mel-image.inc
@@ -11,4 +11,4 @@ inherit core-image image-sanity-incompatible
 
 IMAGE_FEATURES .= "${@bb.utils.contains('COMBINED_FEATURES', 'alsa', ' tools-audio', '', d)}"
 IMAGE_INSTALL += " util-linux-mkfs connman ${@bb.utils.contains("INCOMPATIBLE_LICENSE", "GPLv3", "", "haveged", d)}"
-IMAGE_FEATURES += "splash"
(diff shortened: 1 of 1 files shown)


meta-mel-support/recipes-core/images/mel-image.inc
```diff
@@ -11,4 +11,4 @@ inherit core-image image-sanity-incompatible
 
 IMAGE_FEATURES .= "${@bb.utils.contains('COMBINED_FEATURES', 'alsa', ' tools-audio', '', d)}"
 IMAGE_INSTALL += " util-linux-mkfs connman ${@bb.utils.contains("INCOMPATIBLE_LICENSE", "GPLv3", "", "haveged", d)}"
-IMAGE_FEATURES += "splash"
+IMAGE_FEATURES += "${@bb.utils.contains('EXTRA_IMAGE_FEATURES', 'graphics', 'splash', '', d)}"
```
kergoth: This should probably go based on IMAGE_FEATURES. Directly going based on EXTRA_IMAGE_FEATURES limits options, i.e. we couldn't create an image that's always graphical. I don't intend to do that, though, so this is fine for now, but for future reference.

Noor-Ahsan: @kergoth  core_image-base does not follow this. Do you think we should introduce it. I was thinking not to merge this.

kergoth: Agreed, that's fine.

PR #1245, opened 2019-07-25
glibc-locale: explicitly disable host-user-contaminated

kergoth:
Explicitly disable host-user-contaminated to further work around the
pseudo bug. With pseudo acting up, even if the ownership is correct, it
may well think it is not, so just sidestep the issue until upstream
fixes the root cause.

JIRA: SB-13459


kergoth:
This won't be applied until the manifest change to use current sumo meta-mentor is merged, of course.

PR #1247, opened 2019-08-09
Include multilib libs in production-image, not just development-image

kergoth:


conf/local.conf.sample
```diff
@@ -49,7 +52,7 @@ MACHINE ??= "qemux86"
 # details.
 
 IMAGE_FEATURES_DEVELOPMENT ?= "debug-tweaks codebench-debug ssh-server-openssh tools-profile"
-EXTRA_IMAGE_FEATURES = "${IMAGE_FEATURES_DEVELOPMENT}"
+EXTRA_IMAGE_FEATURES = "${IMAGE_FEATURES_DEVELOPMENT} multilib-runtime"
```
Noor-Ahsan: I think we don't enable it by default. This will require 32bit toolchain on jenkins build also.

kergoth: Only if MULTILIBS has been enabled. The feature includes nothing if no multilib is enabled.

PR #1251, opened 2019-10-28
setup-mel: by default only show the latest version for each machine

kergoth:
Also add a `-a` argument to show all versions.

JIRA: SB-7235

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -11,7 +11,7 @@ ARCHIVE_RELEASE_VERSION = "${DISTRO_VERSION}.${BSP_VERSION}.${PATCH_VERSION}"
 PDK_LICENSE_VERSION_DATE = "20181206"
 
 # Version of the mel-scripts artifact, including setup-mel
-SCRIPTS_VERSION ?= "0"
+SCRIPTS_VERSION ?= "1"
 
 # Default values for BSP and PATCH version, to be redefined in other layers
 BSP_VERSION ?= "0"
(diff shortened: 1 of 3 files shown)


kergoth:
This is currently going against the sumo branch for elm, but would like confirmation on that.

Noor-Ahsan:
I think we introduced setup-mel version. Can you also bump that?

kergoth:
@Noor-Ahsan

Noor-Ahsan:
@abelal you once bump the Version in setup-mel. That was not this one. Can you check which var you bumped.

abelal:
So I believe I bumped the right variable at the wrong place :confused:

I'd suggest that we correct it in the same PR. Also we should probably drop the definition in the recipe altogether as it can be a pitfall...

Noor-Ahsan:
@abelal  can you point which var you modified. share the link.

abelal:
@Noor-Ahsan, here's that: https://github.com/MentorEmbedded/meta-mentor/blob/sumo/meta-mel/mel-support/recipes-core/meta/archive-release.bbappend#L9

kergoth:
Since the previous attempt to bump it didn't actually change the version at all, we shouldn't have to change anything in this PR. Bumping it from 0 to 2 doesn't really help much at this point :) If we weren't going to merge this, then we could bump to 1 in a separate PR, but if we merge this, there isn't much point.

Noor-Ahsan:
Sure. I will merge this.

kergoth:
@Noor-Ahsan, I believe this is good to go.

Noor-Ahsan:
@kergoth  yes. I was just waiting for our current patch release to go out. I want the behavior change from next patch release.

PR #1256, opened 2019-11-12
archive-release: reset SCRIPTS_VERSION and provide a note

abelal:
This is set through the MEL distro config fragment so
it does not make much sense here. Also, can be a bit
confusing in some situations.


abelal:
More of a discussion topic though as to why we have it in the recipe append while we already set it through the distro conf. If there's no need we should probably remove it.

kergoth:
> 
> 
> More of a discussion topic though as to why we have it in the recipe append while we already set it through the distro conf. If there's no need we should probably remove it.

It's there intentionally, for good reason. Not everyone adding a distro layer to bblayers is going to set DISTRO, meaning mel.conf isn't guaranteed to parse. If you're not careful when using undefined variables, you can actually cause the bitbake parsing to fail. The value in the recipe is a fallback default only, to ensure it always exists in some form, regardless of whether mel was set as the distro.

In this particular case it probably wouldn't prevent the parse, but it's a best practice to set defaults for variables being referenced in the recipe or bbclass using it if it's needed and the logic doesn't specifically handle the variable being undefined.

kergoth:
Further, this fragment doesn't actually redefine it, as it uses `?=`, which sets it only if it's not set already.

abelal:
> Further, this fragment doesn't actually redefine it, so the commit message isn't accurate, as it uses `?=`, which sets it only if it's not set already.

That I understand but did not communicate clearly sorry...

abelal:
@kergoth, PR updated. Please re-review.

kergoth:
@Noor-Ahsan , this is approved now.

PR #1257, opened 2019-11-15
setup-mel-builddir: use .machine to set MACHINE when unspecified

kergoth:


scripts/setup-mel-builddir
```diff
@@ -249,16 +249,20 @@ setup_builddir () {
         MACHINE="$1"
     fi
 
+    if [ -z "$MACHINE" ]; then
+        if [ -e "$MELDIR/.machine" ]; then
+            MACHINE="$(cat "$MELDIR/.machine")"
+        elif [ -e "$BUILDDIR/conf/local.conf" ]; then
+            MACHINE=$(sed -n 's/^MACHINE ?*= "\([^"]*\)"/\1/p' $BUILDDIR/conf/local.conf)
```
abelal: A similar action takes place in setup-environment as well (https://github.com/MentorEmbedded/meta-mentor/blob/8f71fff993d6549a6a581aae5dbac573a44eb552/setup-environment#L131) which is used for processing append.MACHINE, should this be extended to that as well?

kergoth: No. setup-environment gets the MACHINE from local.conf after setup-mel-builddir finishes setting up local.conf, which will always be correct. At the time setup-mel-builddir runs, on the other hand, local.conf might or might not even exist at that point.

PR #1264, opened 2020-01-17
Sb 14068: Create mel-examples feature

hashraf91:
Create image feature for mel-examples


Noor-Ahsan:
@hashraf91 what will happend if you build without meta-mentor-private. i think it will break

hashraf91:
@Noor-Ahsan it didn't break, unless you add any layer like meta-mentor-iot which has bbappends of this recipe, else it works fine.

hashraf91:
Closing this PR, image feature mel-examples will be created in meta-mentor-private layer

PR #1265, opened 2020-01-22
Fixes for layer updates

kergoth:


Noor-Ahsan:
@kergoth merging for now. I think we can remove weston image recipe in next PR as we are not using it now.

PR #1278, opened 2020-05-20
master workarounds

kergoth:
- setup-environment: disable meta-tracing until it's fixed
- local.conf.sample: disable external toolchain until the changes are complete


Noor-Ahsan:
we just keep this in mind that we did this workaround :).

kergoth:
> we just keep this in mind that we did this workaround :).

I wonder if it'd be worth adding a mechanism to the release process to fail it out entirely if any comments with a certain string exist in any layers. Ex. FIXME: DO NOT RELEASE WITH THIS WORKAROUND :)

PR #1279, opened 2020-05-20
valgrind: remove, use the oe-core recipe

kergoth:


kergoth:
This fixes the build with current oe-core, upstream has a fix for the gcc 10.1 issue that this copy does not. No need to keep an old copy from elm around anyway, it just got pulled over by the merge.

PR #1281, opened 2020-05-21
optee-os: fix build with gcc 10.1

kergoth:


kergoth:
With this, imx8qxpmek-mel builds again with current layer updates.

PR #1289, opened 2020-06-29
firmware-wireless: move to meta-mentor-bsp

abelal:
The recipe is being utilized for internal purposes only and
is primarily used by BSPs so it will be better placed in the
mentioned layer.


abelal:
This is dependent on https://github.com/MentorEmbedded/meta-mentor-bsp/pull/703 so we don't break anything.

PR #1291, opened 2020-07-06
bb-determine-layers: handle version-specific LAYERDEPENDS

kergoth:
- bb-determine-layers: handle version-specific LAYERDEPENDS
- bb-determine-layers: include version in Layer objects


kergoth:
This is required for https://github.com/MentorEmbedded/mel-manifest/pull/773

PR #1293, opened 2020-07-09
mel: default TCMODE to external-sourcery, not rebuild-libc

kergoth:
There's no need to rebuild it by default given the new toolchain.

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -73,7 +73,7 @@ META_MENTOR_DIR = "${LAYERDIR_mel}/.."
 PATH_prepend = "${META_MENTOR_DIR}/scripts:"
 
 # Default to the external toolchain if available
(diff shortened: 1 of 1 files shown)


Noor-Ahsan:
@kergoth we probably need it. we should not change the behavior that we had previously. That is the guideline provided by marketing.

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Jul 9, 2020, 10:20 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/1293#issuecomment-656251403 "2020-07-09T17:20:14Z - Replied by Github Reply Comments"):
> @kergoth we probably need it. we should not change the behavior that we had previously. That is the guideline provided by marketing.

I'm pretty sure previously we had to rebuild glibc because of the difficulty in releasing new codebench toolchains to roll out cve fixes and the like, but that's now a non-issue, as we'll have the ability to roll out such fixes independently of the rest of the toolchain. That said, we should discuss with product management.

Noor-Ahsan:
Yes that is what I was going to suggest that lets discuss this with marketing. I think another requirement from marketing was user can patch the glibc on their own and rebuild. I get the old MR and we can discuss it with marketing.

PR #1302, opened 2020-08-06
drop encrypted-fs feature from the layer

abelal:
The plan is to implement this through meta-mentor-bsp.


abelal:
related -> https://github.com/MentorEmbedded/meta-mentor-bsp/pull/750

PR #1304, opened 2020-08-10
setup-debian: add initial script

kergoth:
JIRA: SB-15261

### DIFF
--- a/scripts/setup-debian
+++ b/scripts/setup-debian
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+set -u
+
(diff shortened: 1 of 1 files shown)


kergoth:
This is the bare minimum necessary to satisfy the *current* oe sanity test requirements as well as bmap-tools, it doesn't install the lib32 bits right now, since the cb toolchains no longer require that. We'll likely be able to pare down setup-ubuntu a bit at some point in the future. It's possible more may need to be added to this eventually depending on requirements.

PR #1320, opened 2020-09-01
mel.conf: disable sanity check from meta-virtualization

abelal:
Fixes
WARNING: You have included the meta-virtualization layer, but 'virtualization' has not been enabled in your DISTRO_FEATURES. Some bbappend files may not take effect. See the meta-virtualization README for details on enabling virtualization support.

Our support for the layer is implemented a bit differently
as we only pick a handful of packages from it.

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -389,6 +389,11 @@ UPSTREAM_URL_openembedded-core = "https://github.com/openembedded/openembedded-c
 
 FORKED_REPOS ?= "bitbake openembedded-core"
 PUBLIC_REPOS ?= "meta-mentor meta-sourcery meta-tracing ${FORKED_REPOS}"
+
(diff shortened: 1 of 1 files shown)


kergoth:
Have we verified that we aren't building the packages that the distro feature affects?

abelal:
> Have we verified that we aren't building the packages that the distro feature affects?

@kergoth we only use docker from the virtualization-layer atm rather than saying we support the layer or something. The distro feature modifies quite a few packages and I do not think they have been verified in that configuration. If we go that path we'll have to do sanity for all the affected packages whereas I believe the support (docker) has already been tested on all the primary BSPs.

abelal:
Closing... There seems to be an issue with my build env so I was seeing the warning. Our local.conf.sample handles this as https://github.com/MentorEmbedded/meta-mentor/blob/dunfell/conf/local.conf.sample#L100

PR #1323, opened 2020-09-09
local.conf.sample: Reversing commit 9b616.

Noor-Ahsan:
* 9b616 was added for SB-11691. Now SA rework their script.
  See SA-4899 and SA-4905. Now binutils package is not
  needed. So reversing the commit.

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -59,9 +59,6 @@ EXTRA_IMAGE_FEATURES = "${IMAGE_FEATURES_DEVELOPMENT} multilib-runtime"
 IMAGE_FEATURES_PRODUCTION ?= ""
 IMAGE_FEATURES_DISABLED_PRODUCTION ?= "${IMAGE_FEATURES_DEVELOPMENT} ssh-server-dropbear"
 
-# Install binutils with the tools-profile feature, as our tracing scripts need ld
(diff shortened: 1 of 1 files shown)


kergoth:
@Noor-Ahsan , didn't they say they still needed it for another command, not ld?

"Doing a test run on a MEL with removed binutils package shows that the script also requires the readelf utility from that package for reading the Build ID property for executables/libraries. LTTng uses these Build ID for referencing the ELF files.

[~mhussein] - I created SA-4905 for investigating also removing the readelf dependency from the script side."

And SA-4905 is not resolved.

Noor-Ahsan:
@kergoth I checked with Franz he said its fine to remove binutils. 
[1:48 PM] Ahsan Khawaja, Noor (DI SW MG EPT)
    Hello Franz, Are you fine with removing bintuil package from MEL Flex
​[1:53 PM] Polzer, Franz (DI SW MG EPT)
    I‘m still verifying my change. I need to find a CB Installation where I can install the ade into...
​[1:53 PM] Polzer, Franz (DI SW MG EPT)
    The one from
Supporter does not work
​[1:55 PM] Ahsan Khawaja, Noor (DI SW MG EPT)
    As I have to file the oss I will exclude binutils from it
​[1:55 PM] Ahsan Khawaja, Noor (DI SW MG EPT)
    and we can request it later if we have to include it
​[1:56 PM] Polzer, Franz (DI SW MG EPT)
    It should be fine to exclude it

PR #1326, opened 2020-09-11
imx-vpu: correct install paths

kergoth:
Multimedia+graphics+qt5-mel builds fail with an external toolchain due to this recipe always installing to '/usr/lib'.

### DIFF
--- a/meta-mentor-staging/fsl-bsp-release/recipes-bsp/imx-vpu/imx-vpu_%.bbappend
+++ b/meta-mentor-staging/fsl-bsp-release/recipes-bsp/imx-vpu/imx-vpu_%.bbappend
@@ -1,0 +1,11 @@
+do_install_append () {
+    if [ "${libdir}" != "/usr/lib" ]; then
+        install -d "$(dirname "${D}${libdir}")"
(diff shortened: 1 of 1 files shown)


abelal:
The change is correct. However, would it make sense to do this directly in meta-imx? We already have a fork.

Noor-Ahsan:
Merging it atm. Lets move this at meta-imx as we have a fork of it

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Sep 21, 2020, 1:00 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/1326#issuecomment-695963136 "2020-09-21T08:00:28Z - Replied by Github Reply Comments"):
> Merging it atm. Lets move this at meta-imx as we have a fork of it

@Noor-Ahsan @abelal I'm not sure I agree with this. This is a clear bugfix that belongs upstream, which is the definition of meta-mentor-staging. If we have a fork, that's by necessity and ideally would not be necessary, if anything I'd advocate upstreaming anything that belongs in upstream meta-imx to there, and staging it here in the meantime for anything that belongs here.

PR #1327, opened 2020-09-14
os-release: Add more information about MEL Flex OS.

Noor-Ahsan:
* Change the NAME, RETTY_NAME, VERSION and VERSION_ID.
* Add HOME_URL, SUPPORT_URL and BUG_REPORT_URL


Noor-Ahsan:
@kergoth kindly review it under light of https://jira.alm.mentorg.com/browse/SB-15187

Noor-Ahsan:
@kergoth I was thing how we can set ID to flex-os. currently it is set to mel.

meta-mel/recipes-core/os-release/os-release.bbappend
```diff
@@ -0,0 +1,13 @@
+OS_RELEASE_FIELDS = "PRETTY_NAME NAME VERSION_ID VERSION VERSION_CODENAME ID HOME_URL SUPPORT_URL BUG_REPORT_URL BUILD_ID"
+OS_RELEASE_UNQUOTED_FIELDS = "ID VERSION_ID VARIANT_ID"
+
+ID = "${DISTRO}"
```
kergoth, line 4: ID is DISTRO by default already.

meta-mel/recipes-core/os-release/os-release.bbappend
```diff
@@ -0,0 +1,13 @@
...
+NAME = "${OS_NAME}"
```
kergoth, line 5: I'm curious, why is the name Flex OS and not Mentor Embedded Linux Flex OS, is just Flex OS the real product name? Should we use that for DISTRO_NAME?

meta-mel/recipes-core/os-release/os-release.bbappend
```diff
@@ -0,0 +1,13 @@
...
+VERSION = "${DISTRO_VERSION}${@' (%s)' % DISTRO_CODENAME if 'DISTRO_CODENAME' in d else ''} ${YOCTO_RELEASE}"
```
kergoth, line 6: Hmm, it looks like DISTRO_CODENAME actually holds the yocto release for some distros, i.e. the imx and poky distro. In our case, it would probably hold our private name for the release, but if that won't be public anyway, we might just alter our DISTRO_CODENAME to hold the yocto release. It's debatable, but I think we *should* set DISTRO_CODENAME regardless, to either our name or the upstream one. And if we're keeping both, we may want to do the same thing that's done for DISTRO_CODENAME here, rather than putting the parenthesis in the variable contents.

meta-mel/recipes-core/os-release/os-release.bbappend
```diff
@@ -0,0 +1,13 @@
...
+VERSION_ID = "${DISTRO_VERSION}"
+VERSION_CODENAME = "${YOCTO_RELEASE}"
+PRETTY_NAME = "${DISTRO_NAME} ${VERSION}"
+HOME_URL = "https://www.mentor.com/embedded-software/linux/mel-flex-os/"
+SUPPORT_URL = "https://www.mentor.com/support"
+BUG_REPORT_URL = "https://www.mentor.com/support"
```
kergoth, line 12: These should be set in mel.conf, either globally or with _pn-os-release. Actually there's an argument to be made that they should all be there if needed. If they're kept here, they need _mel at a minimum. Still, it's clear the urls don't belong here.

kergoth (changes requested):

meta-mel/recipes-core/os-release/os-release.bbappend
kergoth: We could potentially either set `DISTRO_CODENAME` to `dunfell` or `our-codename (dunfell)`.

meta-mel/recipes-core/os-release/os-release.bbappend
Noor-Ahsan: Sure I will move the urls to mel.conf

meta-mel/recipes-core/os-release/os-release.bbappend
Noor-Ahsan: @kergoth I set the DISTRO_CODENAME in mel.conf. see new commit

meta-mel/recipes-core/os-release/os-release.bbappend
```diff
@@ -0,0 +1,8 @@
+OS_RELEASE_FIELDS = "PRETTY_NAME NAME VERSION_ID VERSION VERSION_CODENAME ID HOME_URL SUPPORT_URL BUG_REPORT_URL BUILD_ID"
+
+ID = "flex-os"
+NAME = "${OS_NAME}"
+VERSION = "${DISTRO_VERSION}${@' (%s)' % DISTRO_CODENAME if 'DISTRO_CODENAME' in d else ''}"
+VERSION_ID = "${DISTRO_VERSION}"
+VERSION_CODENAME = "${@'(%s)' % DISTRO_CODENAME if 'DISTRO_CODENAME' in d else ''}"
+PRETTY_NAME = "${DISTRO_NAME} ${VERSION}"
```
kergoth, line 8: These aren't being defined with _mel or in mel.conf. If changes are made in a bbappend unconditionally, it makes the layer non-yocto-compliant. These need to either be moved to mel.conf or use the override. OS_RELEASE_FIELDS can also be easily appended rather than overridden entirely, no?

meta-mel/conf/distro/include/mel.conf
```diff
@@ -1,8 +1,14 @@
 ## MEL Base Configuration {{{1
 DISTRO = "mel"
 ADE_DISTRO = "mel-flex-os"
-DISTRO_NAME = "Mentor Embedded Linux"
+OS_NAME = "Flex OS"
+GENERIC_NAME = "Mentor Embedded Linux"
+DISTRO_NAME = "${GENERIC_NAME} ${OS_NAME}"
+DISTRO_CODENAME = "dunfell"
```
kergoth, line 7: We may want to create an issue to create a bbclass which adds a sanity check -- make sure DISTRO_CODENAME is listed in LAYERSERIES_CORENAMES. This will ensure that we don't forget to update the DISTRO_CODENAME variable when updating to newer releases.

kergoth (changes requested):

meta-mel/recipes-core/os-release/os-release.bbappend
Noor-Ahsan: Added the _mel pverrride. For OS_RELEASE_FIELDS the order is bit different so I completely override it.

meta-mel/recipes-core/os-release/os-release.bbappend
Noor-Ahsan: I think you got the answer from email.

kergoth (approved):

kergoth:
We can improve this further in the future, but this now alleviates my primary concerns. You'll need to rebase this to resolve the conflicts, though, @Noor-Ahsan.

PR #1328, opened 2020-09-15
virtualization feature

arulcse2008:
Enabling virtualization feature by default.

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -94,7 +94,8 @@ USER_FEATURES += "~3g"
 USER_FEATURES += "~nfc"
 
 # When the meta-virtualization layer is enabled, enable the feature by default
-USER_FEATURES += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'virtualization-layer', 'virtualization', '', d)}"
+USER_FEATURES_append_mel = "${@bb.utils.contains('BBFILE_COLLECTIONS', 'virtualization-layer', ' virtualization', '', d)}"
+EXTRA_IMAGE_FEATURES += "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', 'virtualization', '', d)}"
 
 # How many tasks bitbake should run in parallel
 BB_NUMBER_THREADS ?= "${@int(oe.utils.cpu_count() * 1.5)}"


conf/local.conf.sample
```diff
@@ -93,8 +93,8 @@ USER_FEATURES += "~3g"
 # Disable near field communication support by default
 USER_FEATURES += "~nfc"
 
-# When the meta-virtualization layer is enabled, enable the feature by default
-USER_FEATURES += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'virtualization-layer', 'virtualization', '', d)}"
+# Comment to disable and Uncomment to enable virtualization feature
```
kergoth, line 96: It's already uncommented, I think this is unnecessarily verbose, and also doesn't align with the wording of other entries in this file.

arulcse2008: Updated as per @Noor-Ahsan inputs.

kergoth (changes requested):
This doesn't look bad, but the commit message should be adjusted to align with OE commit message conventions. I.e. `local.conf.sample: enable virtualization image feature based on distro feature` or similar. See https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines

conf/local.conf.sample
```diff
@@ -94,7 +94,8 @@ USER_FEATURES += "~3g"
 USER_FEATURES += "~nfc"
 
 # When the meta-virtualization layer is enabled, enable the feature by default
-USER_FEATURES += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'virtualization-layer', 'virtualization', '', d)}"
+USER_FEATURES_append_mel = "${@bb.utils.contains('BBFILE_COLLECTIONS', 'virtualization-layer', ' virtualization', '', d)}"
```
kergoth, line 97: @Noor-Ahsan It looks like the only place that this distro feature is used other than the meta-virtualization layer is in enabling kernel configs needed. Given the low impact, we might consider just unconditionally enabling the distro feature in mel.conf rather than having yet another change controlled by layer inclusion. Thoughts?

abelal: @kergoth, how would the user be able to disable the feature then? Commenting out image features?

kergoth: Yes, that would prevent inclusion of anything from meta-virtualization. If they really don't want the kernel config, you can still do USER_FEATURES += "~virtualization" to undo what mel.conf adds to DISTRO_FEATURES.

Noor-Ahsan:
closed in favour of https://github.com/MentorEmbedded/meta-mentor/pull/1334 and https://github.com/MentorEmbedded/mel-manifest/pull/786

PR #1335, opened 2020-09-26
Sanitize included image types for installation

kergoth:
JIRA: SB-15569

This avoids shipping tar.bz2 where we can, prefers ext4 to ext3, fixes runqemu so it works at all (doesn't really belong in this pull request, but it's related), avoids shipping buildhistory, and moves to an explicit list of files to include from DEPLOY_DIR_IMAGE, and if a wic image type is enabled, those are the only types we ship.


abelal (approved):
With the limited amount of knowledge that I have around archive-release this looks fine :)

PR #1339, opened 2020-09-30
archive-release-downloads: process git submodules

kergoth:
JIRA: SB-15594, SB-15611


kergoth:
I had to read the gitsm fetcher to nail this down. It seems the upstream archiver bbclass (original mode) and other similar functions are also lacking this handling.

PR #1340, opened 2020-10-05
boost: Set PARALLEL_MAKE to -j 4.

Noor-Ahsan:
* boost takes lots of system resources. Somethimes it give memory error.
  Setting PARALLEL_MAKE to 4 so that it does not cosume all of the
  system resources.


kergoth (changes requested):
PRALLEL_MAKE in the commit subject :)

Noor-Ahsan:
And I also have to add distro overrride

PR #1342, opened 2020-10-05
Add and use codename_is_corename.bbclass

kergoth:
This class implements a sanity check which only makes sense when this DISTRO always defines its DISTRO_CODENAME to match up with the Yocto release codename. If DISTRO_CODENAME is not listed in LAYERSERIES_CORENAMES, we abort.

Example output:
```
This version of mel is incompatible with this version of openembedded-core (code name gatesgarth). Please use matching layer branches or update Mentor Embedded Linux Flex OS.
```

JIRA: SB-15572


abelal (commented):
:+1:

PR #1343, opened 2020-10-05
Move from CODEBENCH_PATH to TOOLCHAINS_PATH

kergoth:
We're moving to a 'toolchains' link and TOOLCHAINS_PATH rather
than 'codebench' and CODEBENCH_PATH.

JIRA: SB-15621


abelal (approved):
Looks good!

PR #1345, opened 2020-10-06
bblayers.mel_utils: process git submodules 

kergoth:
JIRA: SB-15611


kergoth:
This is done the same way as the archive-release fix, by using process_submodules recursively.

PR #1351, opened 2020-10-21
setup-rh: remove codeready repo setup

kergoth:
JIRA: SB-15653


Noor-Ahsan:
Yeah, Bangalore QA RHEL machine was not registered. So waiting on that as @azmik is looking into. his team did not get into this issue as their machine was registered.

PR #1355, opened 2020-11-10
udev-extraconf/mount.sh: ignore lvm in automount

ansar-rasool:
Failure message is shown in boot logs when trying to
mount lvm on emmc while booting from sdcard. This
simply skips lvm while automounting to avoid failure
message in boot logs.

Fixes: https://jira.alm.mentorg.com/browse/SB-15734
Signed-off-by: Ansar Rasool <ansar_rasool@mentor.com>


kergoth (approved):

meta-mentor-staging/recipes-core/udev/udev-extraconf_%.bbappend
```diff
@@ -2,6 +2,7 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 SRC_URI +=  "file://0001-udev-extraconf-mount.sh-add-LABELs-to-mountpoints.patch \
              ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'file://systemd-udevd.service', '', d)} \
              file://0001-udev-extraconf-mount.sh-define-mount-prefix-using-a-.patch \
+             file://0001-udev-extraconf-mount.sh-ignore-lvm-in-automount.patch \
```
abelal, line 5: please add the patch at the end i.e. after the next two patches so the patch series are consolidated.

abelal (approved):
Good to go. Thanks for the work here :)

PR #1356, opened 2020-11-10
udev-extraconf: fix systemd automount issue

ekalemen:
This fixes auto-mounting issue where a manually
unmounted disk would get auto-mounted simply by running
tools such as fdisk from the util-linux package which
generate a BLKRRPART ioctl in order to support any of
their features.

JIRA Ticket: SB-15741

Signed-off-by: ekalemen <eugeny_kalemenev@mentor.com>


abelal (approved):
@ekalemen thanks for the catch.

PR #1357, opened 2020-12-04
udev-extraconf: fix systemd automount issue

ekalemen:
This fixes auto-mounting issue where a manually
unmounted disk would get auto-mounted simply by running
tools such as fdisk from the util-linux package which
generate a BLKRRPART ioctl in order to support any of
their features.

JIRA Ticket: SB-15741

Signed-off-by: ekalemen <eugeny_kalemenev@mentor.com>


kergoth:
In what circumstances *will* it be auto-re-mounted after unmounting after an automount? For external media, if I remove and re-insert after manually unmounting, will it still automount?

ekalemen:
> In what circumstances _will_ it be auto-re-mounted after unmounting after an automount? For external media, if I remove and re-insert after manually unmounting, will it still automount?

Yes, if you re-insert media it will be automounted again. This commit protects from background systemd automount during any operations on disks, such as create, delete partitions with fdisk utility, because some of these utilities use ioctls such as BLKRRPART (block device re-read part) ioctl and in response the kernel generates remove/change/add events if the device is not mounted.

PR #1358, opened 2020-12-04
mel: drop build-id

kergoth:
This was only needed when using an external toolchain which didn't include the build id by default. This is not the case with a Yocto-built toolchain, so we no longer need it.


kergoth:
This fixes an issue where multiple build ids can be set.

abelal (approved):
The change itself is correct but do we have a respective JIRA if it fixes a known issue?

kergoth:
@abelal I don't believe an issue was ever opened for this.

kergoth:
@Noor-Ahsan thoughts?

PR #1372, opened 2021-01-12
Further fixes/adjustments for master

kergoth:


meta-mentor-staging/recipes-support/curl/curl_%.bbappend
```diff
@@ -1,3 +0,0 @@
-EXTRA_OECONF_append_feature-mentor-staging = "\
-    --with-ca-path=${sysconfdir}/ssl/certs \
-"
```
abelal, line 3: Is this something that's handled with the with-ca-bundle switch?
Ref: https://git.openembedded.org/openembedded-core/tree/meta/recipes-support/curl/curl_7.74.0.bb#n58

kergoth: As far as I can tell they're independent, but their defaults are based on `--sysconfdir`. I confirmed that a build of curl has ca-path set correctly with or without the argument, presumably based on the configure.ac's defaults.

meta-mel/recipes-core/meta/meta-environment.bbappend
```diff
@@ -5,3 +5,23 @@ python set_multilib_variants () {
...
+else
+    if [ ! -d "${SDKPATH}" ]; then
+        echo >&2 "Warning: Unable to determine SDK install path from environment setup script location, using default of ${SDKPATH}."
+    fi
+    scriptdir="${SDKPATH}"
+fi
+END
+    sed -e "s#${SDKPATH}#\$scriptdir#g" "$script" >>"${script}.new"
+    mv "${script}.new" "${script}"
+}
```
abelal, line 27: If CB is the only case we need to support this with, wouldn't meta-mentor-private be a better host for this?

kergoth: Hmm that's a good point, thanks. I tend to think of that layer as 'ade', while this technically isn't ade-specific, it does serve the same goal, of supporting CB. I'll move it there once I verify that it's branched.

kergoth: I'm going to take two steps here:
1. Move this bbappend to meta-mentor-private
2. Also modify oe-core's meta-environment to include this, on a branch on oe-core-contrib for a future effort to improve the relocation handling of the sdk, so it doesn't completely fall off my radar

kergoth: Submitted https://github.com/MentorEmbedded/mel-meta-mentor-private/pull/146 for master there and removed the append entirely here.

PR #1376, opened 2021-01-25
wic_bmap: add space between IMAGE_FSTYPES

0xsajjad:


Noor-Ahsan:
@kergoth build is failing so I have to merge it. If there is any better fix then kindly do that.

PR #1378, opened 2021-01-25
sb 16200

kergoth:
- Revert switch to wic-bmap
- wic-bmap.inc: drop toaster reference


kergoth:
The real problem causing the cycle isn't just this, but the fact that the bbclass acted on all images, and our imx-mel-common.inc is doing an IMAGE_FSTYPES_append that also applies to initramfs images, but given we didn't actually have a pressing need to switch the implementation to a class, I think the least risky change at this time is to revert this, drop the toaster reference, and revisit the issue in post-fir.

PR #1379, opened 2021-01-26
setup-mel-builddir: do not depend on mentor-bsp for every machine

abelal:
We have a requirement where upstream Yocto BSPs should be able to
build with MEL out of the box. Sourcing setup-environment at this
point for an upstream machine adds the MEL common BSP layer to
bblayers automatically which is not correct as that particular
machine is not a MEL BSP. However, we do need this for qemu BSPs
as we use the same upstream machine name but create a layer for
our modifications.

### DIFF
--- a/scripts/setup-mel-builddir
+++ b/scripts/setup-mel-builddir
@@ -274,7 +274,7 @@ setup_builddir () {
             fi
         fi
     fi
-    OPTIONALLAYERS="$OPTIONALLAYERS mentor-bsp mentor-bsp-$MACHINE"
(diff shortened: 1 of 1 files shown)


kergoth (changes requested):
I'd suggest unconditionally including mentor-bsp-$MACHINE, as doing so doesn't actually harm anything if the mentor-bsp sublayer doesn't exist, and just remove mentor-bsp (common layer) entirely, rather than hardcoding an assumption of qemu*.

kergoth (approved):

PR #1391, opened 2021-02-24
mel.conf: apply meta-sca imx workaround

kergoth:
JIRA: MELMR-1020


kergoth:
Without this, the task will fail when run against certain meta-imx recipes. Normally the directory would be created in do_rootfs or do_install depending on image vs recipe, but if the recipe flags do_install as noexec, rather than deltask'ing it, the meta-sca task will still run, but the directory it expects won't exist.

PR #1396, opened 2021-02-26
[dunfell]  cve-check.bbclass, cve-update-db-native, cve_check.py: New bits from oe-core for build failure 

Noor-Ahsan:


kergoth (changes requested):
1. Subject line of the PR is gigantic
2. Why remove the bbclass in one commit and then bring it back in another?

Noor-Ahsan:
@kergoth I wanted to tell that a new base has been committed that is why I removed the old class and brought a néw one along other bits as single commit

kergoth (approved):

PR #1400, opened 2021-03-05
mel.conf: do not fetch anti-virus db at compile time in clamav

kergoth:
JIRA: SB-16462

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -234,6 +234,8 @@ PACKAGECONFIG_REMOVE_pn-mesa-demos = "glx"
 # We don't use dracut to build initramfs
 PACKAGECONFIG_REMOVE_pn-plymouth = "initrd"
 
+# Do not fetch the anti-virus db at compile-time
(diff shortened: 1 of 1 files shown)


meta-mel/conf/distro/include/mel.conf
```diff
@@ -234,6 +234,8 @@ PACKAGECONFIG_REMOVE_pn-mesa-demos = "glx"
 # We don't use dracut to build initramfs
 PACKAGECONFIG_REMOVE_pn-plymouth = "initrd"
 
+# Do not fetch the anti-virus db at compile-time
+INSTALL_CLAMAV_CVD = ""
```
abelal, line 238: Would it make more sense to set this using
`INSTALL_CLAMAV_CVD_mel ?= ""`

kergoth: It's defined in the recipe with ?=, and _mel is redundant when it's already being defiend in mel.conf, unless it's necessary. I do like the idea of using ?= to ease overrides from local.conf, though. I'll adjust it.

kergoth: Updated.

PR #1402, opened 2021-03-12
[dunfell] mel-debugging / mel-bsp / inclusion of layers for upstream BSPs

abelal:
- move mel-debugging machine feature from meta-mentor to meta-mentor-bsp
- conditionalize inclusion of machine layer on qemu as other upstream BSPs should be handled differently
- utilize the feature-mel-bsp


abelal:
This is dependent on https://github.com/MentorEmbedded/meta-mentor-bsp/pull/936

kergoth:
@abelal Can you clarify why the conditionalizing of the qemu machine is necessary?

PR #1403, opened 2021-03-17
[dunfell] libsdl2: disable shared memory for native builds

abelal:
libsdl2 currently does not handle the failures that
occur due to shared memory in case of remote hosts
where it should ideally switch to socket based writing
when a failure is seen with shared memory.
Consider a scenario where a qemu build is done with
graphics support on a build machine and then accessed
remotely using ssh. If the remote host's X presents
MIT-SHM as an extension, launching qemu fails with

runqemu - ERROR - Failed to run qemu: X Error:  BadValue
   Request Major code 130 (MIT-SHM)
   Request Minor code 3 ()

This is most often seen when the remote machine doing
ssh is a Ubuntu 20.04. The libsdl2 native are mainly
used for qemu at this time so it is a major usecase.
A report of such a failure was also presented at
https://lists.yoctoproject.org/g/poky/topic/78854857

Fixes: https://jira.alm.mentorg.com/browse/SB-16569

### DIFF
--- a/meta-mentor-staging/recipes-graphics/libsdl2/libsdl2_2.%.bbappend
+++ b/meta-mentor-staging/recipes-graphics/libsdl2/libsdl2_2.%.bbappend
@@ -1,0 +1,1 @@
+CFLAGS_append_class-native = " -DNO_SHARED_MEMORY"


abelal:
As mentioned in the community report link provided in the commit message, another way for us would be to switch to gtk+ from sdl for qemu. I've tested that to be working as well but this seemed like a better choice.

kergoth (approved):

abelal:
Thanks guys. That was _quick_ :)

abelal:
I've submitted this upstream as well but haven't had a response yet: https://lists.openembedded.org/g/openembedded-core/message/149602

abelal:
Our fix got accepted upstream as well https://git.openembedded.org/openembedded-core/commit/meta/recipes-graphics/libsdl2?id=d12306a6331238a25db52816ad74a775c026d7da

PR #1405, opened 2021-03-18
patchinfo: add script

kergoth:
JIRA: ITS-641


kergoth:
This is for use with python clearing due to the need to record what files were modified.

PR #1407, opened 2021-03-22
[dunfell] rootfs.py: fix for disabling run-postinsts on systemd

abelal:
The rootfs generation mechanism removed sysV style scripts
for the run-postinsts when they were unneeded but misssed
the same case for systemd type systems. This caused the
systemd service to run unnecessarily which can cause
uninteded long booting times on platforms that do not have
enough hardware resources.

Fixes: https://jira.alm.mentorg.com/browse/SB-15658

### DIFF
--- a/meta-mentor-staging/lib/oe/__init__.py
+++ b/meta-mentor-staging/lib/oe/__init__.py
@@ -1,0 +1,6 @@
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+from pkgutil import extend_path
+__path__ = extend_path(__path__, __name__)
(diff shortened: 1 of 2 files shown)


abelal:
Submitted upstream as well: https://lists.openembedded.org/g/openembedded-core/message/149764

abelal:
The relevant portion changed here from the upstream is https://github.com/MentorEmbedded/meta-mentor/pull/1407/files#diff-e52845ea96b9a803bee5659853a9f0986fa47d26cef3f3d746cb2d1b65c30748R257

Noor-Ahsan:
@abelal do it in two commits. first one is initial commit and second commit should bring your change.

abelal:
> @abelal do it in two commits. first one is initial commit and second commit should bring your change.

@Noor-Ahsan, done but somehow this isn't getting picked up from meta-mentor-staging....

Noor-Ahsan:
Saw your comment

abelal:
This is now fixed, missed the init module for extending the path. Doing some runtime tests now.

kergoth (approved):
LGTM, nicely done.

PR #1408, opened 2021-03-23
python3: use patchinfo to write python3-patches.txt for release-downloads

kergoth:
This aligns with the behavior of meta-codebench-toolchain for python3 clearing, as suggested by @jeremy-robinson. In this case patchinfo is used, solely because the other script from the other layer is more specific to their requirements rather than general purpose.


jeremy-robinson:
As long as you're generating a summary of changes and including that summary with the product, for example with the Python sources, then you're meeting the legal obligation as I understand it.

kergoth:
@Noor-Ahsan any objections?

PR #1415, opened 2021-04-09
openssh:  fix CVE-2021-28041 by updating the package

abdulkarim17:
This CVE has an impact before version 8.5 so updating
to 8.5p1 version fixes this issue.

Signed-off-by: Karim, Hafiz Abdul <HafizAbdul_Karim@mentor.com>

### DIFF
--- a/meta-mentor-staging/recipes-connectivity/openssh/files/add-test-support-for-busybox.patch
+++ b/meta-mentor-staging/recipes-connectivity/openssh/files/add-test-support-for-busybox.patch
@@ -1,0 +1,47 @@
+Adjust test cases to work with busybox.
+
+- Replace dd parameter "obs" with "bs".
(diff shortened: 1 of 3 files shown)


abelal:
@abdulkarim17 I suggest updating the commit message here.
openssh: upgrade to 8.5p1
then provide details why we are updating and also where we've copied this recipe from.

abelal:
@abdulkarim17, we've a go ahead from management on this i.e. we can switch to 8.5p1 if we are able to use sources from https://cloudflare.cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-8.5p1.tar.gz
Please update the SRC_URI and test with it.

abdulkarim17:
@abelal updated.

abelal (approved):

PR #1417, opened 2021-04-12
[HOLD] mel_utils.py: add source and homepage info for dump-licenses

abelal:
The dump-licenses command additionally processes SRC_URI and
HOMEPAGE for each of the packages. This is a requirement for
the Siemens SVM.
This change will have impact on the OSS jobs that generate
the diffs for OSS submissions as it utilizes the output of
dump-licenses already so modifications to that part should
be considered prior to merging this.


abelal:
Opened the PR so we don't loose the change, we'll keep it open for the near future before we finalize what needs to be done on the OSS approvals job side to remain compliant.

PR #1434, opened 2021-09-10
Add and fix `setup-mel -a` for fir/dunfell

kergoth:
JIRA: SB-17430


kergoth:
It seems the addition of the -a argument was never pulled to newer branches at all, only elm/sumo, and so the fix for it also never went in, so grab both.

Noor-Ahsan:
Send a PR for master as well so that we dont miss it

kergoth:
> Send a PR for master as well so that we dont miss it

We'll likely want to just merge fir to master again before we get far in the next release, I don't think it's worth the hassle of bringing it over one PR at a time.

PR #1435, opened 2021-09-13
Define location for external artifacts and include MCF distro configs

ahsanhussain:


Noor-Ahsan (approved):

meta-mel/conf/distro/include/mel.conf
```diff
@@ -400,6 +400,9 @@ UPSTREAM_URL_openembedded-core = "https://github.com/openembedded/openembedded-c
 
 FORKED_REPOS ?= "bitbake openembedded-core"
 PUBLIC_REPOS ?= "meta-mentor meta-sourcery meta-tracing ${FORKED_REPOS}"
+
+# Define a location for placing external artifacts to be used by the build
+MEL_EXTERNAL_ARTIFACTS = "${TOPDIR}/mel-external-artifacts"
```
abelal, line 405: Again lets use `?=` here to have some flexibility for the future.

ahsanhussain: It's a distro defined location, used by various feature implementations. Why let anyone override that?

abelal: We're not letting anyone override that (not documented or in user facing material) but providing a mechanism for ourselves to override it if and when needed if we have to fulfill a use case where this location might need to be changed.

ahsanhussain: I don't foresee such a scenario, and even if there is, changing it for one feature will break it for the others. Anyway, updating to weak assignment...

abelal (approved):

PR #1437, opened 2021-09-20
[sumo] mel.conf: check for GPLv3 variants for QEMUDEPS

abelal:
INCOMPATIBLE_LICENSE can have any variant of GPLv3 license
names i.e. GPL-3.0 or GPLv3.

Fixes: https://jira.alm.mentorg.com/browse/SB-17451

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -31,7 +31,7 @@ MEL_DEFAULT_EXTRA_RRECOMMENDS = "kernel-module-af-packet"
 DISTRO_EXTRA_RDEPENDS += " ${MEL_DEFAULT_EXTRA_RDEPENDS}"
 DISTRO_EXTRA_RRECOMMENDS += " ${MEL_DEFAULT_EXTRA_RRECOMMENDS}"
 
-QEMUDEPS = "${@bb.utils.contains('INCOMPATIBLE_LICENSE', 'GPL-3.0', '', 'packagegroup-core-device-devel',d)}"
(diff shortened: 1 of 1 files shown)


abelal:
@kergoth I believe this would be required for dunfell etc as well.

kergoth:
Ah, well spotted. Ideally this would use the code in oe.license to handle the mapping to the alternative names, though. license.conf defines that mapping.

kergoth:
Actually: `def incompatible_license_contains(license, truevalue, falsevalue, d)` — looks like there's a function that already does what we want, checks INCOPMATIBLE_LICENSE handling mapping to SPDX names.

abelal:
> incompatible_license_contains

@kergoth I don't believe that's going to help because license.conf just defines spdx license mapping whereas we're looking for a list of license names for GPLv3. That said, I'd still need to update this PR to include GPLv3.0 as well which is also allowed in license string.

kergoth:
> > incompatible_license_contains
> 
> @kergoth I don't believe that's going to help because license.conf just defines spdx license mapping whereas we're looking for a list of license names that correspond to GPL v3 i.e. GPLv3, GPLv3.0 and GPL-3.0. That said, I'd still need to update this PR to include GPLv3.0 as well which is also allowed in license string.

GPL-3.0 *is* the spdx mapping. It maps GPLv3/GPLv3.0 to the canonical license name of GPL-3.0, and also handles a '+' suffix. https://github.com/openembedded/openembedded-core/blob/master/meta/conf/licenses.conf#L42-L49 covers all the variants, including both GPLv3 and GPLv3.0, as well as the + suffixes.

kergoth (approved):
Alright for now, but should be fixed going forward.

PR #1446, opened 2021-12-16
Improve dump-licenses adding --append and --sourceinfo flags

ArsalanHAwan:
This adds --append (-a) and --sourceinfo (-s) flags to mel_utils.py/dump-licenses.

JIRA: SB-18234

See commits for details.

Usage:
--append
$ bitbake-layers dump-licenses development-image
$ bitbake-layers dump-licenses development-image -c populate_sdk -a

--sourceinfo
$ bitbake-layers dump-licenses development-image -s
$ bitbake-layers dump-licenses development-image -c populate_sdk -a -s

### DIFF
--- a/meta-mel/lib/bblayers/mel_utils.py
+++ b/meta-mel/lib/bblayers/mel_utils.py
@@ -231,7 +231,8 @@ class MELUtilsPlugin(LayerPlugin):
 
         fetch_recipes = self._collect_fetch_recipes(args.targets, args.task, depgraph)
 
-        with open(filename, 'w') as f:
+        omode = 'a' if args.append else 'w'
+        with open(filename, omode) as f:
             for recipe in fetch_recipes:
                 fn = depgraph['pn'][recipe]['filename']
                 real_fn, cls, mc = bb.cache.virtualfn2realfn(fn)
@@ -242,6 +243,11 @@ class MELUtilsPlugin(LayerPlugin):
                 pv = data.getVar('PV')
                 lc = data.getVar('LICENSE')
                 f.write('%s,%s,%s\n' % (pn, pv, lc))
+
+        # remove any duplicates added due to append flag
+        uniqlines = set(open(filename).readlines())
+        with open(filename, 'w') as f:
+            f.writelines(uniqlines)
 
     def register_commands(self, sp):
         common = argparse.ArgumentParser(add_help=False)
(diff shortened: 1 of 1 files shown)


ArsalanHAwan:
--extrainfo logic taken from: https://github.com/MentorEmbedded/meta-mentor/pull/1417

ArsalanHAwan:
@kergoth, once reviewed, finalized and merged, should we also merge this into master branch?

kergoth:
> @kergoth, once reviewed, finalized and merged, should we also merge this into master branch?

It will belong in master also, but we'll most likely just merge dun fell to master as a whole in a near future to pick up anything else that's currently missing there.

meta-mel/lib/bblayers/mel_utils.py
```diff
@@ -241,7 +242,22 @@ def do_dump_licenses(self, args):
...
                 lc = data.getVar('LICENSE')
-                f.write('%s,%s,%s\n' % (pn, pv, lc))
+
+                if not args.extrainfo:
+                    f.write('%s,%s,%s\n' % (pn, pv, lc))
+                else:
+                    # For SRC_URI we need to drop the local sources i.e. file:// entries
+                    # then for further filtering we pick up only the initial part of the archive/git
+                    # src i.e. dropping out bitbake specifics like protocol, branch, name etc.
+                    su = " ".join([x.split(';')[0] for x in data.getVar('SRC_URI').split() if not x.startswith('file://')])
```
kergoth, line 252: This isn't a good way to handle this. bb.fetch has methods for parsing the urls, determining the scheme, etc. We should be using that rather than direct string manipulation. You can directly split the list of urls and parse them with `bb.fetch.decodeurl`, or you can construct a fetch object, i.e. `bb.fetch.Fetch(data.getVar('SRC_URI').split(), data)` and access the url data on the fetch object.

kergoth (changes requested):

kergoth:
The concept of these seem sound, though I do wonder if we'd be better off with a more generic 'extra info' argument where you can specify what variables you want added, but this will do.

ArsalanHAwan:
> The concept of these seem sound, though I do wonder if we'd be better off with a more generic 'extra info' argument where you can specify what variables you want added, but that'd make the interface more complex, so this will do.

I'll change that to --sourceinfo because we are extracting SRC_URI and HOMEPAGE only.

meta-mel/lib/bblayers/mel_utils.py
ArsalanHAwan: Done. Thanks!

Noor-Ahsan:
@kergoth can you review this PR.

kergoth:
Reviewed it 19 days ago, checking the latest change to it now. Thanks.

kergoth (approved):

PR #1450, opened 2022-01-10
96boards-tools: remove unused sed from the recipe

ahsanhussain:
This sed is from legacy implementation and no longer applies to the
current disk-resizer script in the 96boards-tools source


ahsanhussain:
Can't use the latest bits on the original repo, which decouples the tool from udev and usees sysfs instead. We're using an older parted version; in contrast the latest changes on the repo use the later version with interactive prompt fixed

Noor-Ahsan:
mention the jira id

ahsanhussain:
This is a general fix. No JIRA ticket has been created against this so didn't put one

PR #1454, opened 2022-01-17
kernel-uboot.bbclass: get rid of compression for arm64 kernel

ahsanhussain:
Maintain a local copy of kernel-uboot.bbclass which overrides the
behavior from the upstream class and removes kernel compression from
arm64 kernel image. This saves up to 1s from slow targets at bootup

JIRA ID: SB-17289

### DIFF
--- a/meta-mel/classes/kernel-uboot.bbclass
+++ b/meta-mel/classes/kernel-uboot.bbclass
@@ -1,0 +1,26 @@
+uboot_prep_kimage() {
+	if [ -e arch/${ARCH}/boot/compressed/vmlinux ]; then
+		vmlinux_path="arch/${ARCH}/boot/compressed/vmlinux"
+		linux_suffix=""
+		linux_comp="none"
+	elif [ -e arch/${ARCH}/boot/vmlinuz.bin ]; then
+		rm -f linux.bin
+		cp -l arch/${ARCH}/boot/vmlinuz.bin linux.bin
+		vmlinux_path=""
+		linux_suffix=""
+		linux_comp="none"
+	else
+		vmlinux_path="vmlinux"
+		linux_suffix=""
+		linux_comp="none"
+	fi
+
+	[ -n "${vmlinux_path}" ] && ${OBJCOPY} -O binary -R .note -R .comment -S "${vmlinux_path}" linux.bin
+
+	if [ "${linux_comp}" != "none" ] ; then
+		gzip -9 linux.bin
+		mv -f "linux.bin${linux_suffix}" linux.bin
+	fi
+
+	echo "${linux_comp}"
+}


ahsanhussain:
No compression means no extra time decompressing the kernel image

```bash
switch to partitions #0, OK
mmc0 is current device

MMC read: dev # 0, block # 8192, count 16 ... 16 blocks read: OK
55478480 bytes read in 3640 ms (14.5 MiB/s)
## Loading kernel from FIT Image at 0c880000 ...
   Using 'conf-system-top.dtb' configuration
   Trying 'kernel-1' kernel subimage
     Description:  Linux kernel
     Created:      2022-01-11  11:49:04 UTC
     Type:         Kernel Image
     Compression:  uncompressed
     Data Start:   0x0c880114
     Data Size:    46184960 Bytes = 44 MiB
     Architecture: AArch64
     OS:           Linux
     Load Address: 0x00000000
     Entry Point:  0x00000000
     Hash algo:    sha256
     Hash value:   6684d4ccf0552a25fd6ac78b8d5d5836ccd776afaffbd49f61427219dab17ac5
   Verifying Hash Integrity ... sha256+ OK
## Loading ramdisk from FIT Image at 0c880000 ...
   Using 'conf-system-top.dtb' configuration
   Trying 'ramdisk-1' ramdisk subimage
     Description:  mel-initramfs-image
     Created:      2022-01-11  11:49:04 UTC
     Type:         RAMDisk Image
     Compression:  uncompressed
     Data Start:   0x0f4995a0
     Data Size:    9235877 Bytes = 8.8 MiB
     Architecture: AArch64
     OS:           Linux
     Load Address: 0x06400000
     Entry Point:  unavailable
     Hash algo:    sha256
     Hash value:   8567d97a0f7691e8d55d875e80a6c6077bff75bb07d322d4933d4859f16ed602
   Verifying Hash Integrity ... sha256+ OK
   Loading ramdisk from 0x0f4995a0 to 0x06400000
## Loading fdt from FIT Image at 0c880000 ...
   Using 'conf-system-top.dtb' configuration
   Trying 'fdt-system-top.dtb' fdt subimage
     Description:  Flattened Device Tree blob
     Created:      2022-01-11  11:49:04 UTC
     Type:         Flat Device Tree
     Compression:  uncompressed
     Data Start:   0x0f48bc24
     Data Size:    55455 Bytes = 54.2 KiB
     Architecture: AArch64
     Load Address: 0x06200000
     Hash algo:    sha256
     Hash value:   830268b8e9a237c3cfc44196ca5af6fc6a4ddf9fe3b3b737f157ca3f48b2afcf
   Verifying Hash Integrity ... sha256+ OK
   Loading fdt from 0x0f48bc24 to 0x06200000
   Booting using the fdt blob at 0x6200000
   Loading Kernel Image
   Using Device Tree in place at 0000000006200000, end 000000000621089e

Starting kernel ...
```

abelal:
Such a change might be worthy of upstreaming if we do it a bit differently using a variable i.e. the user should be able to select whether he needs a compressed kernel image or not. Also, is there a possibility for us to be using the first two if/else clauses to achieve the same?
Even with the answers to above the change would make most sense in staging.
@ahsanhussain thoughts?

kergoth:
> Such a change might be worthy of upstreaming if we do it a bit differently using a variable i.e. the user should be able to select whether he needs a compressed kernel image or not. Also, is there a possibility for us to be using the first two if/else clauses to achieve the same?
> Even with the answers to above the change would make most sense in staging.
> @ahsanhussain thoughts?

For what it's worth, I agree, making this something the user can easily control would be nice.

PR #1456, opened 2022-01-19
A few misc commits from recent POC

kergoth:
This includes a couple fixes and a couple enhancements, none of which will have any effect except in particular cases or when opted in.

- Avoid conflict between `templates` and `images` artifacts in archive-release
- Allow archival of local git repos that have no origin remote
- Include `${VENDOR}.conf` from mel for later vendor config support
- Allow shipment of xlayers.conf to customize default layer inclusion beyond that of the BSP, DISTRO, and MEL default optional layers


kergoth:
Note: ginko/kirkstone, not fir/dunfell.

PR #1457, opened 2022-02-15
mel.conf: Do PAM integration in Flex OS distro

ahsanhussain:
Our images requre PAM to have user slice for user sessions in systemd cgroup
hierarchy. Without this we have login shell running directly under agetty,
causing it to ignore SIGHUP/SIGTERM and causing delays when we try to stop getty

### DIFF
--- a/meta-mel-support/recipes-core/images/mel-image.inc
+++ b/meta-mel-support/recipes-core/images/mel-image.inc
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------------------------------------------------
 
+inherit features_check
+
+REQUIRED_DISTRO_FEATURES = "pam"
+
 # We want a package manager in our base images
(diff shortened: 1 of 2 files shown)


Noor-Ahsan:
@ahsanhussain add the jira issue in the commit message. I am surprised wht was happening in Elm. We did not have pam in Distro in Elm release

Noor-Ahsan:
BTW this is a big change for already releaded bsps.

abelal:
So having PAM would be the way to go IMO as well. However, I'd suggest back-porting https://github.com/MentorEmbedded/meta-mentor/pull/1419 and also asses the runtime impact if any.

abelal:
We're also missing the JIRA ID here.

ArsalanHAwan:
@ahsanhussain I think we should also post the "systemd status" log here that led us to identify the difference in how agetty is being started with and without pam.

ahsanhussain:
Closing this as per the comment on SB-18421

PR #1459, opened 2022-03-18
mel: set https protocol for all github urls

kergoth:
This is a workaround for https://github.blog/2021-09-01-improving-git-protocol-security-github/, to avoid needing to update bitbake to pick up the fix there.


kergoth:
Elm.

abelal (approved):
👍

Noor-Ahsan:
@kergoth do we need this change if we have https://github.com/openembedded/openembedded-core/blob/sumo/meta/classes/mirrors.bbclass#L75

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Mar 24, 2022, 8:14 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/1459#issuecomment-1077740449 "2022-03-24T15:14:14Z - Replied by Github Reply Comments"):
> @kergoth do we need this change if we have https://github.com/openembedded/openembedded-core/blob/sumo/meta/classes/mirrors.bbclass#L75

Not needed, no, it'll fall back to https.

PR #1462, opened 2022-03-29
mel: use common compatible tuning values for arm & aarch64

kergoth:
This avoids unnecessary toolchain component rebuilds and increases reuse and
compatibility amongst compatible machines. Imported, then tweaked to also apply
to aarch64 and our DISTRO.

Imported from https://github.com/96boards/meta-rpb/blob/master/conf/distro/include/arm-defaults.inc
Imported from https://github.com/Angstrom-distribution/meta-angstrom/blob/master/conf/distro/include/arm-defaults.inc

### DIFF
--- a/meta-mel/conf/distro/include/mel-arm-tuning.inc
+++ b/meta-mel/conf/distro/include/mel-arm-tuning.inc
@@ -1,0 +1,28 @@
+# Use common compatible tuning values.
+#
+# This avoids unnecessary toolchain component rebuilds and increases reuse and compatibility amongst
+# compatible machines.
+#
+# Imported from https://github.com/96boards/meta-rpb/blob/master/conf/distro/include/arm-defaults.inc
+# Imported from https://github.com/Angstrom-distribution/meta-angstrom/blob/master/conf/distro/include/arm-defaults.inc
+
+def get_generic_tune(d):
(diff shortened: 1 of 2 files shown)


kergoth:
This applies to ginko/kirkstone, not fir/dunfell.

kergoth:
@jeremy-robinson This adapts existing machines to something close to the existing cb toolchain multilib configs. It's not needed when using the codebench-toolchain- machines, as those set DEFAULTTUNE to such a value explicitly.

Noor-Ahsan:
@kergoth let's first discuss this change in meeting and then merge this. Jeremy's argument was we were moving to internal toolchain for better alingment with upstream but this change is moving away from upstream. So we have to first decide what we want to do here

kergoth:
> @kergoth let's first discuss this change in meeting and then merge this. Jeremy's argument was we were moving to internal toolchain for better alingment with upstream but this change is moving away from upstream. So we have to first decide what we want to do here

I'm fine with discussing in a meeting before merging. To be fair, this does not move away from upstream when it comes to toolchain builds, it only changes the default tuning for the machines to allow for better sharing between them, which is common in multiple Yocto-based distributions, including Angstrom.

PR #1463, opened 2022-03-31
RFC: Remove various bits which support non-GPLv3 builds

kergoth:


kergoth:
Not sure on this one. I don't think we actively support non-gplv3 builds anymore, though.

abelal:
We should probably add @chombourger to this discussion as well. Officially, we do not support GPLv3 excluded builds with the product releases anymore however we know that some of the customers are interested. We should probably remove this from here but create a separate layer with these changes that the services teams could work with and maintain. This would at least give them a starting point.

chombourger:
It's definitely alright for Flex 13. Services team should create a layer to support non-GPLv3 builds that they would only deliver to selected customers/projects

conf/local.conf.sample
```diff
@@ -96,21 +96,6 @@ BB_NUMBER_THREADS ?= "${@int(oe.utils.cpu_count() * 1.5)}"
 # How many jobs 'make' should run in parallel when compiling
 PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count() * 2}"
 
-# Uncomment to exclude GPLv3 software from the build
-#INCOMPATIBLE_LICENSE = "GPL-3.0* LGPL-3.0*"
```
Noor-Ahsan, line 100: @chombourger it was decided that the option will remain there in local.conf to build with non GPLV3

kergoth: @Noor-Ahsan Ah, if that's the case then this PR should be dropped, as it won't be usable without the rest.

PR #1466, opened 2022-04-13
Remove the no-longer-supported 'blocks' mechanism

kergoth:


meta-mel/recipes-core/packagegroups/packagegroup-base.bbappend
```diff
@@ -1,8 +1,3 @@
-BLUEZ ?= "bluez5"
-VIRTUAL-RUNTIME_bluetooth-stack ?= "${BLUEZ}"
-VIRTUAL-RUNTIME_bluetooth-hw-support ?= ""
-
-RDEPENDS:packagegroup-base-bluetooth:mel = "${VIRTUAL-RUNTIME_bluetooth-stack} ${@bb.utils.contains('DISTRO_FEATURES', 'pulseaudio', 'pulseaudio pulseaudio-server', '', d)}"
-RRECOMMENDS:packagegroup-base-bluetooth:mel = "${VIRTUAL-RUNTIME_bluetooth-hw-support}"
+RDEPENDS:packagegroup-base-bluetooth:append:mel = "${@bb.utils.contains('DISTRO_FEATURES', 'pulseaudio', 'pulseaudio pulseaudio-server', '', d)}"
```
abelal, line 1: why aren't we adding these to the packagegroup-base-bluetooth and not the packagegroup-base-distro?

kergoth: I honestly don't remember, just left that line alone as it was since this PR was to drop the blocks. I'll check git history.

abelal: We should probably merge this as is then and come back to that later?

abelal (approved):

meta-mel/recipes-core/packagegroups/packagegroup-base.bbappend
kergoth: Probably. I'm intending a line-by-line, file-by-file review of every layer as we prepare for ginko when we have the resources available to do so.

PR #1467, opened 2022-04-13
WIP: mel.conf improvements and cleanup

kergoth:


kergoth:
@Noor-Ahsan @abelal Please review

meta-mel/conf/distro/mel.conf
```diff
@@ -187,6 +187,14 @@ require conf/distro/include/yocto-uninative.inc
 
 INHERIT += "uninative"
 ## }}}1
+## Mechanisms provided for user customization {{{1
+# Support USER_FEATURES
+INHERIT += "user_features"
```
abelal, line 153: What's the relevance with user features here?

meta-mel/conf/distro/mel.conf
```diff
@@ -232,9 +240,6 @@ BUILDHISTORY_COMMIT ?= "1"
 # Copy the image license manifest into DEPLOY_DIR_IMAGE
 INHERIT += "deploy-license-manifest"
 
-# Support USER_FEATURES
-INHERIT += "user_features"
```
abelal, line 236: ah ok so it was moved but in the same commit?

meta-mel/conf/distro/mel.conf
```diff
@@ -175,11 +175,6 @@ LINUX_VERSION_TAG ??= "v${@'.'.join('${LINUX_VERSION}'.split('.')[:2])}"
 LINUX_VERSION ??= ""
 BB_GIT_SHALLOW_REVS:pn-linux-mel ??= "${@'${LINUX_VERSION_TAG}' if '${LINUX_VERSION_TAG}' and '${LINUX_VERSION}' else ''}"
 
-# Export path variables into the devshell for convenience
-OE_TOPDIR = "${TOPDIR}"
-OE_WORKDIR = "${WORKDIR}"
-OE_TERMINAL_EXPORTS += "OE_TOPDIR OE_WORKDIR COREBASE"
```
abelal, line 183: Shouldn't we keep it if it's pending upstream?

meta-mel/conf/distro/mel.conf
```diff
@@ -177,6 +177,7 @@ BB_GIT_SHALLOW_REVS:pn-linux-mel ??= "${@'${LINUX_VERSION_TAG}' if '${LINUX_VERS
 
 require conf/distro/include/no-static-libs.inc
 require conf/distro/include/yocto-uninative.inc
+require conf/distro/include/security_flags.inc
```
abelal, line 147: This looks a bit strict to me. Are we ready for this? Various packages might have issues due to the strictness of flags that this implies. We should at least keep a buffer in our estimates for the next product release if we're to follow this.

meta-mel/conf/distro/mel.conf
```diff
@@ -78,9 +78,6 @@ MELDIR ?= "${COREBASE}/.."
 # Use a local PR server by default
 PRSERV_HOST ?= "localhost:0"
 
-# Default to ipk packaging
-PACKAGE_CLASSES ?= "package_ipk"
```
abelal, line 87: What if the default we're relying on moves? Is this something we have picked up consciously in the past? If so, then we should keep this otherwise we're ok.

meta-mel/conf/distro/mel.conf
```diff
@@ -112,8 +112,8 @@ do_image_wic[recrdeptask] += "do_populate_lic"
 # on do_populate_lic
 do_image_complete[recrdeptask] += "do_populate_lic"
 
-# Quadruple the normal. 'du' is not a good way to really see how much
-# space will be needed and fails badly as the fs size grows.
+# Quadruple the normal extra space in an image. 'du' is not a good way to really
+# see how much space will be needed and fails badly as the fs size grows.
 IMAGE_ROOTFS_EXTRA_SPACE = "40960"
```
abelal, line 117: Now that we do wic images we've an ability add extra space to the partition from the wic definition. We should probably think over doing this from a single point or else we'll be adding additional space multiple times.

meta-mel/conf/distro/mel.conf
```diff
@@ -184,9 +184,8 @@ PACKAGECONFIG:append:pn-pulseaudio = " autospawn-for-root"
 PACKAGECONFIG_REMOVE:pn-plymouth = "initrd"
 ## }}}1
 ## Inherits {{{1
-# We want information about package and image contents
+# We want information and history about build output
 INHERIT += "buildhistory"
-BUILDHISTORY_COMMIT ?= "1"
```
abelal, line 230: Again, should we rely on the upstream default or keep it as a distro decision?

meta-mel/conf/distro/mel.conf
```diff
@@ -320,11 +320,10 @@ DL_LICENSE_INCLUDE ?= "*"
 INHERIT += "archive-release-downloads"
 
 # For our forks, also point an 'upstream' remote at upstream
-UPSTREAM_URL = ""
-UPSTREAM_URL_bitbake = "https://github.com/openembedded/bitbake"
-UPSTREAM_URL_openembedded-core = "https://github.com/openembedded/openembedded-core"
+UPSTREAM_URL_meta-virtualization = "git://git.yoctoproject.org/meta-virtualization"
+FORKED_REPOS ?= "meta-virtualization"
```
abelal, line 324: hmmmmm so in order to do this right the BSP layers should append to this appropriately? We do fork vendor repos sometimes.

kergoth: Yeah, mel.conf has the baseline for flex os, but bsps can add to it when we fork those layers. This only applies to public layers, and it just arranges for git remotes to exist in the installed git repositories so the customer can easily fetch both the fork and upstream, as a usability improvement. That said, if one is missed, it's certainly  non-critical, which is how this got out of date in the first place.

meta-mel/conf/distro/mel.conf
kergoth: Yes, just rearranged because we have a number of cases where we provide mechanisms to make certain things easier from local.conf for the customer.

meta-mel/conf/distro/mel.conf
kergoth: Yeah, I'll move that one into a separate RFC/WIP PR until it gets merged.

meta-mel/conf/distro/mel.conf
kergoth: Yeah, I mainly wanted to start the discussion on this. I can open a draft MR. It aligns with poky and other open source distros and I think is a step in the right direction, but as you say, could have impact if anything breaks as a result.

meta-mel/conf/distro/mel.conf
kergoth: Hmm, perhaps. Will wic automatically resize the filesystem image it's pointed to to match the partition size, or does that rely on the first boot partition resizing?

meta-mel/conf/distro/mel.conf
kergoth: Just moved from a different part of the file to consolidate user customization hooks/mechanisms.

kergoth:
Thanks for the thoughts, @abelal, I'll update the PR!

meta-mel/conf/distro/mel.conf
abelal: wic does that automatically if we use the --extra-space or --overhead option. Ref: https://docs.yoctoproject.org/ref-manual/kickstart.html#command-part-or-partition and does not depend on the first boot resizing.

meta-mel/conf/distro/mel.conf
kergoth: Rebased and moved this to a different branch.

meta-mel/conf/distro/mel.conf
kergoth: Started an email thread and dropped the commit from this PR

meta-mel/conf/distro/mel.conf
kergoth: Dropped this commit, in this case we want to be explicit, you're right

meta-mel/conf/distro/mel.conf
kergoth: Opened an issue in the repo as a reminder

meta-mel/conf/distro/mel.conf
kergoth: In this case I don't think it's necessary, since we don't explicitly require this particular feature for our product.

PR #1468, opened 2022-04-19
archive-release: improve reproducibility and usability for releases

kergoth:
- Avoid duplication of pack files between releases and updates
- Avoid creating new commits for public layers
- Ship history for public layers, don't make them shallow
- Improve the commit message when we do create commits

[Screencast](https://asciinema.org/a/pcXiBFSPVeUzj5wTnXuJDFEal)


kergoth:
The fact that we didn't ship history on public layers doesn't make us look great, even though it doesn't actually prevent you from running `git fetch --unshallow` to get the history, and the space savings aren't huge for layers, so just ship the history.

PR #1473, opened 2022-04-22
mel: only support the tunings aligned with the external toolchain

kergoth:
This aborts the MEL Flex build if the user is attempting to build a configuration which isn't one of the supported codebench toolchain configurations. This will keep internal toolchain builds aligned with external toolchain builds. It also means that attempts to build unsupported architectures like mips will fail at this time.


kergoth:
This aborts the MEL Flex build if the user is attempting to build a configuration which isn't one of the supported codebench toolchain configurations. This will keep internal toolchain builds aligned with external toolchain builds. It also means that attempts to build unsupported architectures like mips will fail at this time.

kergoth:
This list is hardcoded at this time, but given it's dependent upon the available machine files in another layer, I think there's no cleaner method at this time.

jsm28 (commented):
Seems reasonable to me.

PR #1474, opened 2022-05-05
WIP: initial jobs/threads heuristic

kergoth:
JIRA: MELMR-1124

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -91,10 +91,10 @@ USER_FEATURES += "~3g"
 USER_FEATURES += "~nfc"
 
 # How many tasks bitbake should run in parallel
-BB_NUMBER_THREADS ?= "${@int(oe.utils.cpu_count() * 1.5)}"
+#BB_NUMBER_THREADS ?= "${@int(oe.utils.cpu_count())}"
 
 # How many jobs 'make' should run in parallel when compiling
-PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count() * 2}"
+#PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count()}"
 
 # Uncomment to exclude GPLv3 software from the build
 #INCOMPATIBLE_LICENSE = "GPL-3.0* LGPL-3.0*"
--- a/meta-mel/conf/distro/include/jobs-heuristic.inc
+++ b/meta-mel/conf/distro/include/jobs-heuristic.inc
@@ -1,0 +1,95 @@
+# Set bitbake threads and make jobs based upon available memory per CPU core.
+# Based on https://elinux.org/images/d/d4/Goulart.pdf, a talk done by Mario Domenech
+# Goulart and Mikko Rapeli from BMW CAR IT GMBH.
+#
+# MEM_PER_CPU_MINIMUM,THREADS,JOBS,ACTION
+JOBS_HEURISTIC_THRESHOLDS ?= "\
+    7.9,cpus,cpus*1.5, \
+    3.9,cpus,cpus, \
+    1.9,cpus//2,cpus//2,WARN \
+    0,,,FATAL \
+"
+
+# This function postpones warnings to reduce duplication, and postpones failures
+# to avoid breaking bitbake -e.
+python jobs_heuristic() {
+    """Set bitbake threads and make jobs based upon available memory per CPU core."""
+    cpus = oe.utils.cpu_count()
+    try:
+        mem_total = mem_total_gigs()
+    except Exception:
+        d.setVar('JOBS_HEURISTIC_WARNING',
+                 'Failed to determine system total memory for threads and jobs heuristic.')
+        return
+    mem_per_cpu = mem_total / cpus
+
+    if cpus == 1:
+        threads, jobs = 1, 1
+    else:
+        for entry in d.getVar('JOBS_HEURISTIC_THRESHOLDS').split():
+            split = entry.split(',')
+            try:
+                required_mem_per_cpu, threads, jobs, action = split
+            except ValueError:
+                d.setVar('JOBS_HEURISTIC_ERROR',
+                         'Invalid entry in JOBS_HEURISTIC_THRESHOLDS: %s. Expected: MEM_PER_CPU_MINIMUM,THREADS,JOBS,ACTION' % entry)
+                return
+
+            if mem_per_cpu >= float(required_mem_per_cpu):
+                if action:
+                    if action == 'FATAL':
+                        d.setVar('JOBS_HEURISTIC_ERROR',
+                                 'Not enough memory per CPU core (~%s Gb) in this system, exiting.' % int(mem_per_cpu))
+                        return
+                    elif action == 'WARN':
+                        d.setVar('JOBS_HEURISTIC_WARNING',
+                                 'Low memory per CPU core (~%s Gb) in this system, parallelization has been scaled down.' % int(mem_per_cpu))
+
+                try:
+                    threads = int(eval(threads)) if threads else 1
+                except Exception as exc:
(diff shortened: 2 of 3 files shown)


meta-mel/conf/distro/include/jobs-heuristic.inc
```diff
@@ -0,0 +1,96 @@
...
+                    d.setVar('JOBS_HEURISTIC_ERROR',
+                             'Failed to evaluate jobs (%s) in JOBS_HEURISTIC_THRESHOLDS entry %s: %s' % (
+                                jobs, entry, exc))
+                    return
+                break
+        else:
+            return
+
+    bb.debug(1, 'Memory per CPU core is %s, adjusting BB_NUMBER_THREADS=%s, PARALLEL_MAKE=%s' % (mem_per_cpu, threads, jobs))
+    if not d.getVar('BB_NUMBER_THREADS'):
```
abelal, line 68: isn't this set from the bitbake.conf already in the default case?

meta-mel/conf/distro/include/jobs-heuristic.inc
```diff
@@ -0,0 +1,96 @@
...
+        d.setVar('BB_NUMBER_THREADS', str(max(threads, 1)))
+    if not d.getVar('PARALLEL_MAKE'):
```
abelal, line 70: and so is this?

meta-mel/conf/distro/include/jobs-heuristic.inc
kergoth: Yes, good point, I'll update it.

meta-mel/conf/distro/include/jobs-heuristic.inc
kergoth: Same.

meta-mel/conf/distro/include/jobs-heuristic.inc
```diff
@@ -0,0 +1,95 @@
...
+                                jobs, entry, exc))
+                    return
+                break
+        else:
+            return
+
+    bb.debug(1, 'Memory per CPU core is %s, adjusting BB_NUMBER_THREADS to %s and PARALLEL_MAKE to %s' %
+                  (mem_per_cpu, threads, jobs))
+    d.setVar('BB_NUMBER_THREADS', str(max(threads, 1)))
+    d.setVar('PARALLEL_MAKE', '-j %s' % max(jobs, 1))
```
abelal, line 70: Wouldn't this override whatever we set in local.conf?

conf/local.conf.sample
```diff
@@ -91,10 +91,10 @@ USER_FEATURES += "~3g"
 USER_FEATURES += "~nfc"
 
 # How many tasks bitbake should run in parallel
-BB_NUMBER_THREADS ?= "${@int(oe.utils.cpu_count() * 1.5)}"
+# BB_NUMBER_THREADS ?= "${@int(oe.utils.cpu_count())}"
 
 # How many jobs 'make' should run in parallel when compiling
-PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count() * 2}"
+# PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count()}"
```
abelal, line 97: Drop the space after the # in these variable commenting cases? I guess that's how we do it for such stuff.

meta-mel/conf/distro/include/jobs-heuristic.inc
kergoth: Updated to set intermediate variables which are then referenced by mel.conf with ?=. This ensures that local.conf would be preferred, as it's parsed prior to mel.conf, which is parsed prior to the defaults in bitbake.conf.

conf/local.conf.sample
kergoth: Fixed, thanks

meta-mel/conf/distro/include/jobs-heuristic.inc
```diff
@@ -0,0 +1,95 @@
+# Set bitbake threads and make jobs based upon available memory per CPU core.
+# Based on https://elinux.org/images/d/d4/Goulart.pdf, a talk done by Mario Domenech
```
abelal, line 2: So we've implemented the algo as per page 50 of the mentioned study? If yes, that looks a bit different to me. Comments to follow on the exact scenarios.

meta-mel/conf/distro/include/jobs-heuristic.inc
```diff
@@ -0,0 +1,95 @@
...
+# Goulart and Mikko Rapeli from BMW CAR IT GMBH.
+#
+# MEM_PER_CPU_MINIMUM,THREADS,JOBS,ACTION
+JOBS_HEURISTIC_THRESHOLDS ?= "\
+    7.9,cpus,cpus*1.5, \
```
abelal, line 7: for 8G or greater per cpu the study uses threads=cpus pmake=cpus, we've pmake=cpus*1.5 here.

meta-mel/conf/distro/include/jobs-heuristic.inc
```diff
@@ -0,0 +1,95 @@
...
+    3.9,cpus,cpus, \
```
abelal, line 8: for greater than 4G/cpu the study uses threads=cpus pmake=cpus/2, we've pmake=cpus

meta-mel/conf/distro/include/jobs-heuristic.inc
```diff
@@ -0,0 +1,95 @@
...
+    1.9,cpus//2,cpus//2,WARN \
```
abelal, line 9: aligned with the study here

meta-mel/conf/distro/mel.conf
```diff
@@ -47,6 +47,15 @@ WARN_TO_ERROR_QA ??= ""
 WARN_QA:remove = "${WARN_TO_ERROR_QA}"
 ERROR_QA:append = " ${WARN_TO_ERROR_QA}"
 
+# Adapt default threads and jobs based on available memory
+require conf/distro/include/jobs-heuristic.inc
+BB_NUMBER_THREADS ?= "${BB_NUMBER_THREADS_HEURISTIC}"
+PARALLEL_MAKE ?= "${PARALLEL_MAKME_HEURISTIC}"
```
abelal, line 53: typo here
PARALLEL_MAK**M**E_HEURISTIC

kergoth:
@abelal I think I tweaked it based on our current default values which are more aggressive than the defaults in bitbake.conf, we'll have to decide if we want to continue our more aggressive defaults, which tend to hit problems in constrained environments, or switch to upstream more conservative defaults and align with the study. I'm inclined to the latter, in which case it should be changed based on your comments. Thoughts, @Noor-Ahsan ?

abelal:
> @abelal I think I tweaked it based on our current default values which are more aggressive than the defaults in bitbake.conf, we'll have to decide if we want to continue our more aggressive defaults, which tend to hit problems in constrained environments, or switch to upstream more conservative defaults and align with the study. I'm inclined to the latter, in which case it should be changed based on your comments. Thoughts, @Noor-Ahsan ?

I'd agree to this.

Noor-Ahsan:
> > @abelal I think I tweaked it based on our current default values which are more aggressive than the defaults in bitbake.conf, we'll have to decide if we want to continue our more aggressive defaults, which tend to hit problems in constrained environments, or switch to upstream more conservative defaults and align with the study. I'm inclined to the latter, in which case it should be changed based on your comments. Thoughts, @Noor-Ahsan ?
> 
> I'd agree to this.

I would agree to this. our defautls is based on study that we did at early stage of the product. After that we did not update it.

kergoth:
Migrated to https://github.com/MentorEmbedded/meta-sokol-flex/pull/13

PR #1477, opened 2022-05-16
[dunfell]: archive-release: drop false initialization of SUBLAYERS_INDIVIDUAL_ONLY variable

ansar-rasool:
This drops false initializations of SUBLAYERS_INDIVIDUAL_ONLY variable

JIRA-ID: PSP-12481

Signed-off-by: ansar-rasool <ansar_rasool@mentor.com>

### DIFF
--- a/meta-mel-support/recipes-core/meta/archive-release.bb
+++ b/meta-mel-support/recipes-core/meta/archive-release.bb
@@ -41,15 +41,6 @@ def configured_update_layers(d):
             update_layers.add(layer)
     return ' '.join(sorted(update_layers))
 
-def configured_mx6_layers(d):
-    """Return the mx6 layers to be archived as individual tarballs"""
-    mx6_layers = set()
-    for layer in d.getVar('BBLAYERS').split():
-        basename = os.path.basename(layer)
-        if 'mx6' in basename:
-            mx6_layers.add(layer)
-    return ' '.join(sorted(mx6_layers))
-
 # Sub-layers to archive individually, rather than grabbing the entire
 # repository they're in
 def layers_by_name(d, *layers):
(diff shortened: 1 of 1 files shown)


abelal (approved):

Noor-Ahsan:
@kergoth @ansar-rasool why we are hitting this only for singgo layer? why other layers are not impacted during jenkins build when archive-layers task is executed?

ansar-rasool:
Removed default addition of mentor-bsp layer in SUBLAYERS_INDIVIDUAL_ONLY var, as now, each layer adds itself explicitly to this var.

meta-mel-support/recipes-core/meta/archive-release.bb
```diff
@@ -58,7 +58,7 @@ def layers_by_name(d, *layers):
         if v:
             yield v
 
-SUBLAYERS_INDIVIDUAL_ONLY ?= "${@configured_mx6_layers(d)}"
```
abelal: if the function (configured_mx6_layers) isn't being used elsewhere in the file please drop it too.

ansar-rasool: dropped

kergoth:
[**@ansar-rasool**](https://github.com/ansar-rasool) commented on [May 16, 2022, 2:51 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/1477#issuecomment-1127462160 "2022-05-16T09:51:02Z - Replied by Github Reply Comments"):
> Upon investigation, i found out that due to ${@' '.join(layers_by_name(d, 'mentor-bsp', 'mentor-bsp-${MACHINE}'))}   this line in
> https://github.com/MentorEmbedded/meta-mentor/blob/dunfell/meta-mel-support/recipes-core/meta/archive-release.bb#L61
>  archive_mel_layers failure issue is happening for singo layer only. 
> As meta-mentor-singgo-aitronx is a BSP layer it adds itself into the SUBLAYERS_INDIVIDUAL_ONLY var and causes the issue,
> on the other hand, our meta-mentor-bsp layer does have sublayers in it and works fine. Also, all other layers that we have in metadata don't have a layer name with "mentor-bsp" suffix, which is why issue doesn't appear on those layers.

Well spotted, indeed, the issue is a layer outside of meta-mentor-bsp using the mentor-bsp suffix and breaking that assumption in meta-mentor, that should be dropped, but I still think archive-release shouldn't break if a layer adds itself to sublayers when it isn't a sublayer :)

PR #1478, opened 2022-05-16
siejir 6302

kergoth:
- scripts: add git wrapper from oe-core for safe directory fix
- mel: add meta-mentor-staging/scripts to the PATH


Noor-Ahsan:
We have workaround in https://github.com/MentorEmbedded/meta-mentor/pull/1479. we don't need this PR atm.

PR #1479, opened 2022-05-18
mel: work around safe dir issue with METADATA_*

kergoth:
With recent versions of git, attempting to operate on a git repository owned by
one user as another user will fail. As our builds run under pseudo in some
contexts, this fails. This issue in tasks can be worked around with a git
wrapper script, and is worked around upstream, but that does not resolve the
issue with metadata_scm.bbclass, as it isn't run with the correct PATH. It's a
non-issue upstream as they already made this change to force immediate expansion
of these variables in the right context, so apply that here also.

JIRA: SB-19344

### DIFF
--- a/meta-mel/conf/distro/include/mel.conf
+++ b/meta-mel/conf/distro/include/mel.conf
@@ -350,6 +350,12 @@ MACHINE_HWCODECS = ""
 
 # Work around missing vardep bug in bitbake
 sstate_stage_all[vardeps] += "sstate_stage_dirs"
+
+# Work around git safe directory issue with METADATA_*, by forcing early expansion
+# while in a build user context.
+require classes/metadata_scm.bbclass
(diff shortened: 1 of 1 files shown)


Noor-Ahsan:
@kergoth should we revert this when we move to new oe-core in our next update release. where we will be getting git wrapper script from oe-core/scripts folder?

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [May 19, 2022, 3:43 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/1479#issuecomment-1131532720 "2022-05-19T10:43:57Z - Replied by Github Reply Comments"):
> @kergoth should we revert this when we move to new oe-core in our next update release. where we will be getting git wrapper script from oe-core/scripts folder?

That depends on what version of oe-core we update to, but perhaps.

PR #1480, opened 2022-05-26
mel_utils: catch SRC_URI items missing from variable history

kergoth:
These are now printed with the layer name of the main recipe file, as we were
unable to determine the correct layer to map it to. A warning is also printed
for this case.

JIRA: PSP-12481


meta-mel/lib/bblayers/mel_utils.py
```diff
@@ -296,10 +302,11 @@ def do_dump_licenses(self, args):
     def register_commands(self, sp):
         common = argparse.ArgumentParser(add_help=False)
         common.add_argument('--mirrortarballs', '-m', help='show existing mirror tarball paths rather than git clone paths', action='store_true')
-        common.add_argument('--task', '-c', help='specify the task to use when not expicitly specified for a target (default: BB_DEFAULT_TASK or "build")')
+        common.add_argument('--targets-only', '-t', help='only operate against the specified targets, not their dependencies', action='store_true')
```
abelal, line 305: If I understand this correctly I think we need a separate commit for this?

meta-mel/lib/bblayers/mel_utils.py
```diff
@@ -147,11 +152,13 @@ def _collect_fetch_recipes(self, targets, ctask, depgraph):
...
 
+        tdepends = depgraph['tdepends']
         for task, taskdeps in tdepends.items():
             for dep in taskdeps:
                 deprecipe, deptask = dep.rsplit('.', 1)
-                if deptask == 'do_fetch':
-                    fetch_recipes.add(deprecipe)
+                if include_depends or deprecipe in targets:
+                    if deptask == 'do_fetch':
+                        fetch_recipes.add(deprecipe)
```
abelal, line 161: this part relates to the targets_only functionality, right?

abelal:
This certainly looks like the way to go unless this is handled accurately in varhistory/bitbake. The only concern/comment I have here is that it looks like this should be split into two commits.

kergoth:
[**@abelal**](https://github.com/abelal) commented on [May 27, 2022, 12:18 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/1480#issuecomment-1139353162 "2022-05-27T07:18:27Z - Replied by Github Reply Comments"):
> This certainly looks like the way to go unless this is handled accurately in varhistory/bitbake. The only concern/comment I have here is that it looks like this should be split into two commits.
> Thanks a lot @kergoth  :)

Oops, you're right, accidentally mixed those up, thanks! And no problem

abelal (approved):
:+1:

PR #1482, opened 2022-06-08
mel-selftest: add distro for use with oe-selftest

kergoth:
It's useful to be able to run oe-selftest with mel as well as with
nodistro, but elements of our configuration are incompatible, so add
a convenience distro to remove those elements.


kergoth:
Closing this due to rebranding, but also reconsidering this as a distro feature instead.

PR #1483, opened 2022-06-09
96boards-tools: update to latest revision

ahsanhussain:
This has some improvements like exact size/blocks calculation and
non-interactive parted command

JIRA-ID: SB-19428


Noor-Ahsan:
the non-interactive parted command was already there in commit https://github.com/96boards/96boards-tools/commit/a999d655417bb19ce8a476ff8c811e957748b661. we were already on that commit

ahsanhussain:
@Noor-Ahsan this commit only pipes the arguments, and doesn't make parted non-interactive.
The thing is, required args can change based on the system. The latest one uses a different command for which we don't have to specify the percentage of the partition table to resize the FS to

ahsanhussain:
Regardless, we need to test this on another BSP before merging

PR #1487, opened 2022-06-16
Allow bblayers.conf to affect our bb-determine-layers script

kergoth:
This aligns with the behavior of bitbake itself and allows for us to use this mechanism to pull in additional layers whenever a vendor BSP layer is included, so we can apply necessary integration changes.

JIRA: MELMR-1099

Example usage in our bblayers.conf.sample:
```bitbake
# BSP Integration Layers
LAYERRECOMMENDS_meta-polarfire-soc-yocto-bsp:append = " mentor-vendor-polarfire-soc"
```


kergoth:
Migrated to meta-sokol-flex.

PR #1490, opened 2022-07-06
bb-determine-layers: use common data for all parsed layer.conf files 

kergoth:
This allows for variable references amongst the layers, particularly from a layer to its dependencies.

JIRA: SB-19738

### DIFF
--- a/scripts/bb-determine-layers
+++ b/scripts/bb-determine-layers
@@ -178,6 +178,12 @@ class LayerError(Exception):
     pass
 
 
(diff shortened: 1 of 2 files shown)


abelal (approved):
LVGTM!

abelal:
This will also be need on kirkstone.

Noor-Ahsan:
@kergoth Jira IDs are missing in commit message

kergoth:
This was migrated to meta-sokol-flex and merged.

PR #1496, opened 2022-07-19
Merge from dunfell (kirkstone)

kergoth:


kergoth:
Work-in-progress. Once the server I was working on becomes accessible again, I'll finish the last of the updates to the files copied and overridden from oe-core.

kergoth:
Merging so as to proceed with rebranding, can revisit to fix any issues we spot later.

PR #1500, opened 2022-07-28
udev: fix for systemd mounting timeout issue for dmsetup encrypted pa…

paw3142:
…rtition

Source: https://bugs.gentoo.org/706434

Signed-off-by: Amy Fong <Amy_Fong@mentor.com>


abelal:
This should ideally be done through an lvm2 which actually brings in the udev rules for dm.

PR #1505, opened 2022-11-17
[dunfell] lttng-modules: fix patch failure for qemu machines

tarmeh2r:
This moves the append file and related patches to common vendor layers as these patches are not applicable for qemu machines for two reasons:

    1. For qemu, lttng-modules v2.11.6 is built while for other BSPs v2.12.2 is built.
    2. We do not push LTS updates for linux-yocto (built for qemu) so there's
    no need for these patches in the first place.

Renaming lttng-modules_%.bbappend to lttng-modules_2.12.2.bbappend would also not work as it would give "no recipe available for lttng-modules_2.12.2.bbappend" since its recipe is coming from common vendor layers under meta-mentor-bsp.

This resolves the following error:
Applying patch 0001-fix-random-tracepoints-removed-in-stable-kernels.patch patching file probes/Kbuild
Hunk #1 FAILED at 187.
1 out of 1 hunk FAILED -- rejects in file probes/Kbuild Patch 0001-fix-random-tracepoints-removed-in-stable-kernels.patch does not apply (enforce with -f)

JIRA-ID: SB-20840

Signed-off-by: Tariq Mehmood <tariq_mehmood@mentor.com>


tarmeh2r:
Patch 0001-fix-random-tracepoints-removed-in-stable-kernels.patch was merged in oe-core via https://github.com/openembedded/openembedded-core/commit/9f301f5563df868626d624c2d0781dae1b81a4c0#diff-44f5dea8c750e2d35ef2c60624f7d2c4db8a8cedfed4494e31d7f33fec90a1b8 which is why when lttng-modules recipe from oe-core is built (for qemu machines) then it fails to apply the patch from meta-mel

PR #1508, opened 2023-02-16
udev-extraconf:mount.sh: fix a umount issue

ansar-rasool:
Only touching /tmp/.automount-$name is not good enough, it must contain the mount name, otherwise umount could not get the path from it.

(cherry picked from commit ccea69032329f3ba43c727d9eb71b1d063b89824)


abelal:
missing JIRA id

ansar-rasool:
Updated.

abelal (approved):

PR #1513, opened 2023-08-01
conf: distro.conf: WHITELIST coreutils

ssnaaz:
WHITELIST coreutils for selinux IMAGE_FEATURES when INCOMPATIBLE_LICENSE = "GPLv3" is set

It was leading to the following failure during boot: [FAILED] Failed to start SELinux init for /dev service loading. See 'systemctl status selinux-labeldev.service' for details. [FAILED] Failed to start SELinux init service loading. See 'systemctl status selinux-init.service' for details.

JIRA-ID: SB-22107


kergoth (changes requested):
Whitelisting like this is potentially dangerous, legally, to a customer. They're ending up with gplv3 components when they explicitly requested *not* to have gplv3 components. Further, you're doing it in the distro, which isn't user-visible.

tarmeh2r:
Closing this PR as the issue has been resolved in https://github.com/MentorEmbedded/meta-mentor/pull/1514

PR #1514, opened 2023-08-10
busybox: enable chcon in non-gplv3 builds when selinux is enabled

tarmeh2r:
Coreutils is not built in non-gplv3 builds due to which chcon is not available on rootfs causing the selinux service to fail. Therefore, utilize chcon from busybox when selinux feature is enabled in non-gplv3 builds.

JIRA: SB-22107


meta-mel/recipes-core/busybox/busybox_%.bbappend
```diff
@@ -1,5 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/busybox:"
 
+DEPENDS += "${@bb.utils.contains('INCOMPATIBLE_LICENSE', 'GPLv3', bb.utils.contains('IMAGE_FEATURES', 'security-selinux', 'libselinux', '', d), '', d)}"
```
abelal, line 3: checking image features this way is not the way to go. Please see how we utilize feature-selinux overrides.

meta-mel/recipes-core/busybox/busybox_%.bbappend
```diff
@@ -8,4 +10,5 @@ SRC_URI_append_mel = "\
     file://fancy-head.cfg \
 	file://pidof.cfg \
 	file://top.cfg \
+    ${@bb.utils.contains('INCOMPATIBLE_LICENSE', 'GPLv3', bb.utils.contains('IMAGE_FEATURES', 'security-selinux', 'file://chcon.cfg', '', d), '', d)} \
```
abelal: I'd suggest setting up a variable above on this basis and then adding that to SRC_URI. That would be more flexible. Also the indentation looks a bit off.

meta-mel/recipes-core/busybox/busybox_%.bbappend
tarmeh2r: I believe we don't have feature overrides in Elm release

meta-mel/recipes-core/busybox/busybox_%.bbappend
tarmeh2r: Done. Please have a look

kergoth: Just a note that checking image features outside the image is highly specific to how we do things in mel/sokol-flex, upstream would never accept it. Not an issue for this PR, this is fine, but just so people know.. from a yocto perspective image features can be set entirely in the image, it's just us that set it globally more often than not :)

kergoth:
Actually, how much does enabling chcon add to the busybox binary size? If the impact is trivial is might be cleaner to just unconditionally enable it.

tarmeh2r:
Enabling chcon will not add much to binary size, but we are adding the dependency on libselinux since it is required by busybox when we enable config_selinux which, I think, we would want to avoid unless selinux feature is enabled

meta-mel/recipes-core/busybox/busybox_%.bbappend
tarmeh2r: Thanks for sharing this. It is good to know the proper way to use it.

kergoth:
[**@tarmeh2r**](https://github.com/tarmeh2r) commented on [Aug 10, 2023, 8:23 AM MST](https://github.com/MentorEmbedded/meta-mentor/pull/1514#issuecomment-1673439769 "2023-08-10T15:23:07Z - Replied by Github Reply Comments"):
> Enabling chcon will not add much to binary size, but we are adding the dependency on libselinux since it is required by busybox when we enable config_selinux which, I think, we would want to avoid unless selinux feature is enabled
> 
> _Originally posted by @tarmeh2r in https://github.com/MentorEmbedded/meta-mentor/issues/1514#issuecomment-1673439769_

Fair enough, thanks for indulging my curiosity! Thanks for your work on this.

meta-mel/recipes-core/busybox/busybox_%.bbappend
abelal: Ah my bad.... definitely aging!

abelal (approved):

PR #1518, opened 2024-03-04
cve_check: Remove bbclass and python script to resolve spdx license generation issue

pragow0k:
JIRA_ID: SB-23235


bhargavdas:
The commit message should be something related to cve_check class and script.
IMHO, since we are not supporting spdx generation in product this should be not be mentioned in the commits as well.

CC: @Noor-Ahsan

Noor-Ahsan:
I think we are fine mentioning spdx license. @pragow0k  kindly change your commit summary to somthing like

cve_check: Remove bbclass and python script to resolve spdx license generation issue

In detailed log you can mention that our stuff was outdated so removing these to get new fixed from upstream.

bhargavdas (approved):

